PR #131, opened 2022-08-18
Add sentry agent key to sentry event as label

seunghyupoh3517:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/131)


harjotgill:
Why do we need this?

PR #193, opened 2022-08-25
Prom metrics lifecycle in OTEL layer

DariaKunoichi:
Ref: GH-5352

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/193)

### DIFF
--- a/pkg/otel/components.go
+++ b/pkg/otel/components.go
@@ -16,6 +16,7 @@ import (
 	"go.opentelemetry.io/collector/processor/batchprocessor"
 	"go.opentelemetry.io/collector/processor/memorylimiterprocessor"
 	"go.opentelemetry.io/collector/receiver/otlpreceiver"
+	"go.uber.org/fx"
 	"go.uber.org/multierr"
 
 	"github.com/fluxninja/aperture/pkg/entitycache"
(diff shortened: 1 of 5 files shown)


krdln (changes requested):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/193)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @DariaKunoichi and @krdln)

*[a discussion](https://reviewable.io/reviews/fluxninja/aperture/193#-NAduauI3p_H7pLyy_SI:-NAduauI3p_H7pLyy_SJ:b-m53ngt) (no related file):*
I think there's a disconnect between https://github.com/fluxninja/cloud/issues/5352 and this implementation.

The issue was about "too many per-policy metrics in high policy churn", but in impl, I see that we have a **constant** number of metrics. So.. I'm not sure how that helps. I feel the issue was about somehow removing obsolete MetricVecs that get created here https://github.com/fluxninja/aperture/blob/60524bfb33c53a19d1736b7b1a710e3269b865ee/pkg/otelcollector/metricsprocessor/processor.go#L252 (GetMetricWith(labels) creates a new histogram on first use).

As for the PR itself, I see no value in unregistering the metrics from promRegistry if we're going to "stop the world" anyway.

___

DariaKunoichi:
This PR will not clean up workload metrics - have to prepare part2 with moving registration/cleaning to ResponseMetricsRegistry

PR #197, opened 2022-08-26
update k8s deployment readme and add kind cluster script

slayer321:
Add Installation script for 3 nodes kind cluster.
Update aperture installation Readme for  k8s.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/197)

### DIFF
--- a/scripts/aperture_dev_cluster.sh
+++ b/scripts/aperture_dev_cluster.sh
@@ -1,0 +1,186 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Values selected due to minikube using them by default
+# If this is changed, update entries in GCP
+readonly cluster_docker_network="kind"
+git_root="$(git rev-parse --show-toplevel)"
+readonly git_root="${git_root}"
+readonly kind_node_version='kindest/node:v1.23.3'
+readonly registry_name='docker-registry'
+readonly registry_port='5001' # Docker registry usually uses port 5000, but MacOS Monterey uses it for AirPlay server
+readonly registry_image='registry:2'
+
+setup() {
+  create_registry
+  case "${1:-}" in
+    kind) setup_kind_cluster;;
+    *) printf 'Unknown cluster provider - "%s"\n' "${1:-}"; return 1;;
+  esac
+  document_registry
+}
+
+teardown() {
+  case "${1:-}" in
+    kind) teardown_kind_cluster;;
+    *) printf 'Unknown cluster provider - "%s"\n' "${1:-}"; return 1;;
+  esac
+}
+
+main() {
+  local -r subcommand="${1:-}"
+  local -r cluster_provider="${2:-${NINJA_DEV_CLUSTER_PROVIDER:-kind}}"
+  # We need to handle it earlier, before cluster_provider check logic kicks in
+  if [ "${subcommand:-}" = "help" ]; then
+    print_help help
+    return
+  fi
+
+  # Unused for now, but we might use it soon-ish if we add rancher support for MacOS folks
+  if [ -z "${cluster_provider:-}" ]; then
+    printf 'ERROR: You need to specify what cluster provider to use.\n' >&2
+    printf 'See "help" subcommand for details.\n' >&2
+    return 1
+  fi
+
+  if ! command -v "${cluster_provider}" &>/dev/null; then
+    # The provider command is not available
+    printf 'ERROR: The selected provider is not installed. Install it by going to:\n' >&2
+    case "${cluster_provider}" in
+      kind) print 'https://kind.sigs.k8s.io/docs/user/quick-start/#installation\n' >&2;;
+      *) printf 'Unknown cluster provider - "%s"\n' "${cluster_provider}";;
+    esac
+    return 1
+  fi
+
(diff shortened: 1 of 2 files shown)


scripts/aperture_dev_cluster.sh
```diff
@@ -0,0 +1,229 @@
...
+
+set -euo pipefail
+
+# Values selected due to minikube using them by default
+# If this is changed, update entries in GCP
+readonly cluster_subnet_prefix=192.168.49
+readonly cluster_subnet_mask=24
+readonly cluster_docker_network="ninja-cluster"
+readonly cluster_subnet="${cluster_subnet_prefix}.0/${cluster_subnet_mask}"
+readonly cluster_gateway="${cluster_subnet_prefix}.1"
```
sbienkow-ninja, line 11: Since we don't have UI here (And don't configure contour at all), I'd honestly drop all of the docker-network related steps. It'll make it significantly smaller and simpler.

sbienkow-ninja (changes requested):

scripts/aperture_dev_cluster.sh
slayer321: We also have the internal docker registry which is been created and connected to the docker network . So if we are removing the docker-network steps . We will also need to remove the registry code right ? 
Which I think we will need for the 3 nodes setup

sbienkow-ninja: We can keep the registry - it just needs to be hooked into the kind network then. Docs: https://kind.sigs.k8s.io/docs/user/local-registry/#create-a-cluster-and-registry
Simply replace `cluster_docker_network` value with `kind` and leave the registry code as-is, it should work just fine.

slayer321: Yeah, I made the changes and was testing But I saw that @seunghyupoh3517 , did a [PR](https://github.com/fluxninja/aperture/pull/259) where is mention how can we create 3 nodes cluster using `kind-config.yaml` file .  
So is this PR still relevant as we can create the 3 node cluster using this command
```
kind create cluster --name aperture-playground --config kind-config.yaml
```

seunghyupoh3517: Yes, I needed to three nodes cluster for the rate limiter testing so I created the simple configuration file while discussing with @tanveergill. With the `kind-config.yaml`, we now have a control plane and two worker nodes kind cluster.

slayer321: So is this PR still relevant . Do we need the script to start the three node cluster or just kind-cluster.yaml file works ?

seunghyupoh3517: kind-cluster.yaml file works to start the three node cluster but I haven't taken a look into rest of the codes in the scripts as well as the script in the cloud repo so I cannot tell if it's still needed.

tanveergill: @fluxninja/devops know more about the requirements of 
cluster for Tilt environment. Please make a call on whether we need a script vs bare kind config will do

slayer321:
As @sbienkow-ninja  mention that we don't need the docker-network code in this script. So I have removed that.
@seunghyupoh3517 added the `kind-config.yaml` file which will create the simple 3 node kind cluster , which he added for test purpose. So after having discussion with @hdkshingala , we decided to continue with this script as we will require docker-registry which is been created in this script . 
cc @tanveergill

slayer321:
Hey, @sbienkow-ninja can  you review the PR. I have made the expected changes .

sbienkow-ninja (approved):
:lgtm:

> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/197#-:-NBD0HnJ13GDURnXFmMS:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @fluxninja)

___
*[`scripts/aperture_dev_cluster.sh` line 11 at r1](https://reviewable.io/reviews/fluxninja/aperture/197#-NAPHvF_a3v9w5X-DYSU-r1-11:-NBD0EzW8P_dI_M9dbUK:bg7igkv) ([raw file](https://github.com/fluxninja/aperture/blob/745f7a80711fbdedd8a4949039bf7a5df725b77b/scripts/aperture_dev_cluster.sh#L11)):*
<details><summary><i>Previously, tanveergill (Tanveer Gill) wrote…</i></summary><blockquote>

@fluxninja/devops know more about the requirements of 
cluster for Tilt environment. Please make a call on whether we need a script vs bare kind config will do
</blockquote></details>

:lgtm:

slayer321:
Just saw that @harjotgill have recently made some [changes](https://github.com/fluxninja/aperture/blob/main/playground/README.md#prerequisites---k8s-cluster-bootstrap) in the README.md file.
 `kind-config.yaml`  is used to launch the 3 node kind cluster. 
Can you confirm are we going with script to just the  `kind-config.yaml`  file.
cc @harjotgill , @sbienkow-ninja

hdkshingala (approved):

PR #251, opened 2022-08-30
Fix CLA rendering and improve aperture-operator handling

sbienkow-ninja:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/251)

### DIFF
--- a/manifests/charts/aperture-operator/values.yaml
+++ b/manifests/charts/aperture-operator/values.yaml
@@ -3,7 +3,7 @@
 global:
   ## @param global.imageRegistry Global Docker image registry
   ##
-  imageRegistry: ""
+  imageRegistry: ~
   ## @param global.imagePullSecrets Global Docker registry secret names as an array
   ## E.g.
   ## imagePullSecrets:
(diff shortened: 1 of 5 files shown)


manifests/charts/aperture-operator/values.yaml
```diff
@@ -355,9 +355,9 @@ agent:
   ## @param agent.image.pullSecrets Agent image pull secrets
   ##
   image:
-    registry: ~
-    repository: ~
-    tag: ~
+    registry: docker.io/fluxninja
```
hdkshingala, line 358: This values are configured as default in the operator so we can skip them for Agent and Controller section.

sbienkow-ninja: Oh, ok. When I was testing it tomorrow, it tried to download image from the wrong registry, but it's possible that it was just the operator. I wonder if it'd be worth generating the `values.yaml` here based on the operator defaults - so they are easier to view by users?

hdkshingala: If we add all the defaults here, the generated YAML for CR will be long and user may feel that the CR needs all the values always. Thinking this, I kept only the required ones. What's your thoughts on that?

sbienkow-ninja: That's also a good point. I'm fine either way.

hdkshingala: Okay.. let's continue with the minimum required ones for now and will add all if we see confusion around that later.

PR #274, opened 2022-08-31
Add policies map as first lvl field in heartbeats

DariaKunoichi:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/274)


pkg/policies/controlplane/policy-factory.go
```diff
@@ -118,6 +130,7 @@ func (factory *policyFactory) provideControllerPolicyFxOptions(
 
 	// save policy wrapper proto in status registry
 	reg.Child("policy_config").SetStatus(status.NewStatus(&wrapperMessage, nil))
+	factory.policyWrapperList = append(factory.policyWrapperList, &wrapperMessage)
```
harjotgill, line 133: where are we removing policies when they are removed from the system?

harjotgill: let's make this a map that maps policy name to config and we remove this value when the policy is removed.

DariaKunoichi: Factory seems to be only spawning policies and it does not track removing them. The only place I found when I could have attached the removing is in the etcd notifier for Remove action. But that would be very ugly solution.
I think I'll close this PR because it's not a priority.

DariaKunoichi:
Closing as described above.

pkg/policies/controlplane/policy-factory.go
harjotgill: You  need to add a tracker for policies. it's simply a map.

DariaKunoichi: I've added a map in policy factory instead of list, that's easy, but what tracker are you talking about - new etcd notifier inside policy factory reacting to Remove? wouldn't that be ugly?

harjotgill: just make a normal map in policy factory and add remove entries from it when policies load/unload.

DariaKunoichi: I will try to explain myself on slack.

PR #297, opened 2022-09-01
Prom metrics lifecycle in OTEL layer

DariaKunoichi:
For managing policy related metrics lifetime
Closes: GH-5352

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/297)


DariaKunoichi:
I don't know where exactly to put cleaning of workload related metrics - in concurrency limiter?

pkg/policies/dataplane/response-metrics-registry.go
```diff
@@ -1,10 +1,146 @@
...
+					rm.fluxmeterHistogramVec = are.ExistingCollector.(*prometheus.HistogramVec)
+				} else {
+					errs = multierr.Append(errs, err)
+				}
+			}
+
+			return errs
+		},
+		OnStop: func(_ context.Context) error {
+			unregistered := registry.Unregister(rm.workloadLatencySummaryVec)
```
tanveergill, line 83: Should write deinit code as reverse of init code as a best practice I.E. De register order should be reverse of Register

pkg/policies/dataplane/response-metrics-registry.go
```diff
@@ -1,10 +1,146 @@
...
+	}, []string{
+		metrics.PolicyNameLabel,
+		metrics.PolicyHashLabel,
+		metrics.ComponentIndexLabel,
+		metrics.DecisionTypeLabel,
+		metrics.WorkloadIndexLabel,
+	})
+
+	// FLUXMETER METRICS
+	rm.fluxmeterHistogramVec = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
tanveergill, line 44: Each FluxMeter histogram has different buckets. Vector can represent only one set of buckets. Therefore, cannot use a single Vector to represent all FluxMeter metrics since buckets would change based on FluxMeter

pkg/policies/dataplane/resources/fluxmeter/flux-meter.go
```diff
@@ -138,52 +139,26 @@ func (fluxMeterFactory *fluxMeterFactory) newFluxMeterOptions(
...
-				metrics.StatusCodeLabel,
-				metrics.FeatureStatusLabel,
-			})
-			// Register metric with Prometheus
-			err := prometheusRegistry.Register(fluxMeter.histMetricVec)
-			if err != nil {
-				log.Error().Err(err).Msgf("Failed to register metric %+v with Prometheus registry", fluxMeter.histMetricVec)
-				return err
-			}
-
```
tanveergill, line 162: @DariaKunoichi,
See how a new FluxMeter metrics Vector is created for each FluxMeter mini Fx app. Need to restore the same semantics in order to support different histogram buckets per FluxMeter

pkg/policies/dataplane/resources/fluxmeter/flux-meter.go
```diff
@@ -138,52 +139,26 @@ func (fluxMeterFactory *fluxMeterFactory) newFluxMeterOptions(
...
 			if err != nil {
 				log.Error().Err(err).Msgf("Failed to unregister FluxMeter %s with EngineAPI", fluxMeter.fluxMeterName)
-				errMulti = multierr.Append(errMulti, err)
 			}
 
-			// Unregister metric with Prometheus
-			unregistered := prometheusRegistry.Unregister(fluxMeter.histMetricVec)
-			if !unregistered {
-				log.Error().Err(err).Msgf("Failed to unregister metric %+v with Prometheus registry", fluxMeter.histMetricVec)
-			}
```
tanveergill, line 184: @DariaKunoichi,
Please restore this code. It is already properly handling the lifecycle for FluxMeter metrics. Need lifecylce handling for only the Workload metrics

tanveergill (approved):
Can merge after addressing comments and testing in playground

tanveergill (commented):
@DariaKunoichi,
There is a race condition in workload metrics creation/deletion. Metrics creation is happening in OTEL pipeline. Whereas metric deletion is happening at Policy deletion. This can lead to Workload metric getting re-generated if an access log for a workload is received after policy deletion

Also, the way workload metrics are being created is pretty unsafe. We are not checking the existence of a policy and workload before creating a metric dynamically from the logs.

Refer how we solve these issues in FluxMeter. Essentially the metrics have to be owned by the ConcurrencyLimiter in order to have a clear lifecycle

PR #323, opened 2022-09-05
Response duration, FluxMeter attribute key, miscellaneous fixes

tanveergill:
# What
* Use [response duration](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators) instead of duration for calculating latencies
* Change `workload_latency_ms` from histogram to summary since histograms are not required for auto tokens calculation
* Flux Meter metric can be switched to any attribute key in access log or span attributes, defaults to duration_millis.
* Set FluxMeter buckets in histogram metric. Buckets were not getting set since [this change](https://github.com/fluxninja/aperture/pull/162).
* Remove stray workload metric being published for rate limiters.
* Fix metric parsing from OTEL attributes
  * Always parse metrics as float64 not int64
  * Handle parsing from string, int or float type attributes
  * `-` is treated as zero when parsing duration field from Envoy
* Rollup refactor
  * Specify separate rollups for different sources.
  * Can add different ones for Envoy vs other service meshes in the future.
  * Also, specify how to treat an attribute as missing from source per source.
* Sanitizer log/span input and output of OTEL pipeline
  * Include list before enrichment
  * Exclude list after metrics processor

# Why
* Track latency to first byte of response instead of last byte because network delays can be quite unpredictable and don't indicate server overload

This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/323)


tanveergill:
Before:
gradient-algo-with-e2e-latency

After:
gradient-algo-with-response-latency-fixed

harjotgill (approved):

PR #333, opened 2022-09-05
Add presentation layer for compiled circuit

jmichalak-fluxninja:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/333)


pkg/policies/controlplane/circuit.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+	for _, l := range links {
+		g.Edge(clusters[l.Source.ComponentID].Node(l.Source.PortName),
+			clusters[l.Target.ComponentID].Node(l.Target.PortName)).Attr("label", l.SignalName)
+	}
+	return g.String()
+}
+
+type Port struct {
```
harjotgill, line 107: Should we move these to protobuf?
cc: @tanveergill

tanveergill: That’s a good idea. We were initially thinking GraphQL would capture this schema. But since policy graph is now just a blob from GraphQL perspective we are missing it’s schema definition. Protobuf is probably the right format to define its schema and frontend can use the same proto definition for parsing it

kwapik (commented):
LGTM, but move structs to proto as requested.

> Reviewed 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/333#-:-NBMc26Z7sjIlH5h4z8V:b-mxa1e2)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @harjotgill, @kwapik, and @tanveergill)

jmichalak-fluxninja:
Moved to proto.

kwapik (approved):
> Reviewed 3 of 5 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/333)* status: all files reviewed, 1 unresolved discussion (waiting on @harjotgill and @tanveergill)

tanveergill (approved):

PR #344, opened 2022-09-06
Fixed issues observed in operator with fluxninja plugin

hdkshingala:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/344)

### DIFF
--- a/pkg/watchdog/policy.go
+++ b/pkg/watchdog/policy.go
@@ -20,15 +20,19 @@ import (
 // +kubebuilder:object:generate=true
 type WatchdogConfig struct {
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={initial_delay:"0s"}
 	Job jobs.JobConfig `json:"job"`
 
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={watermarks_policy:{watermarks:{0.50,0.75,0.80,0.85,0.90,0.95,0.99}}}
 	CGroup WatchdogPolicyType `json:"cgroup"`
 
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={watermarks_policy:{watermarks:{0.50,0.75,0.80,0.85,0.90,0.95,0.99}}}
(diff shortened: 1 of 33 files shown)


plugins/service/aperture-plugin-fluxninja/pluginconfig/pluginconfig.go
```diff
@@ -27,6 +27,7 @@ type FluxNinjaPluginConfig struct {
 	APIKey string `json:"api_key,omitempty"`
 	// Client configuration.
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={http:{network_keep_alive:"30s"}}
```
harjotgill, line 30: Why do we have to provide this default over here?

pkg/watchdog/policy.go
```diff
@@ -20,15 +20,19 @@ import (
...
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={watermarks_policy:{watermarks:{0.50,0.75,0.80,0.85,0.90,0.95,0.99}}}
 	CGroup WatchdogPolicyType `json:"cgroup"`
 
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={watermarks_policy:{watermarks:{0.50,0.75,0.80,0.85,0.90,0.95,0.99}}}
 	System WatchdogPolicyType `json:"system"`
 
 	//+kubebuilder:validation:Optional
+	//+kubebuilder:default:={limit:268435456,watermarks_policy:{watermarks:{0.50,0.75,0.80,0.85,0.90,0.95,0.99}}}
```
harjotgill, line 35: This is too much code repetion. Are we sure we are using kubebuilder correctly?

harjotgill: we should probably just omitempty in this case

harjotgill:
We can close this.

hdkshingala:
These issues were handled as part of https://github.com/fluxninja/aperture/pull/364.

Closing this PR.

PR #359, opened 2022-09-07
Replace tolerance with slope in Gradient Controller

### LINKED ISSUE #182: Modify Gradient Controller to take slope parameter instead of gradient
Currently,
$$gradient = \frac{setpoint \cdot tolerance}{signal}$$

Proposal:

$$
gradient = \left(\frac{signal}{setpoint}\right)^{slope}
$$

Rationale:
* current `tolerance` parameter is just pre-multiplying setpoint and could be done outside using math `mul` component,
* slope parameter can be used to fine-tune the "aggresiveness" / slope of the gradient,
* current design allows only to use the gradient controller for cases where signal is positively correlated with control variable.

Migration:
* add a mul node if $\text{tolerance} \neq 1$
* set the slope to $-1$.

TODO
* [x] code
* [x] proto docs
* [ ] example policies using this controller (wherever they are commited)

As discussed with @tanveergill

krdln:
Resolves #182

See #182 for more details.

Migration:
* add a mul node if $\text{tolerance} \neq 1$ (pre-multiply the setpoint by _tolerance_ constant),
* set the slope to $-1$.

[Something like here.](https://github.com/fluxninja/aperture/pull/359/files#diff-8f270640fa6c26ce2176a7050e6617cccee5569dc2126cbc379794118ea7ee4c)

TODO:
- [x] jsonnet & blueprints (this PR & https://github.com/fluxninja/aperture-blueprints/pull/25)
- [ ] test on cluster

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/359)


harjotgill (approved):

docs/gen/policies/policy.md
```diff
@@ -1151,15 +1146,48 @@ Some changes are expected in the near future:
...
+
+- Use $\text{slope} < 0$ if signal and control variable are _positively_
+  correlated (eg. Per-pod CPU usage and total concurrency).
+- Use $\text{slope} > 0$ if signal and control variable are _negatively_
+  correlated (eg. Per-pod CPU usage and number of pods).
+
+:::note
+You need to set _negative_ slope for a _positive_ correlation, as you're
+describing the _action_ which controller should make when the signal
+icreases.
```
sbienkow-ninja, line 1174:
```suggestion
increases.
```

sbienkow-ninja:
@krdln Just FYI: https://github.com/fluxninja/aperture-blueprints/pull/25
Also, I see in one place you set `slope` as a number (where previously `tolerance` was a string), while in others both are numbers.
I understand that the values that are read are coerced/deserialized to the target type regardless of the input type?

pkg/policies/controlplane/validator_test.go
```diff
@@ -126,15 +126,30 @@ data:
...
+        - constant:
+            value: "1.1"
+            out_ports:
+              output:
+                signal_name: "EMA_SETPOINT_MULTIPLIER"
+        - arithmetic_combinator:
+            operator: "mul"
+            in_ports:
+              lhs:
+                signal_name: "LATENCY"
```
tanveergill, line 138: Need to multiply with `LATENCY_EMA` instead of `LATENCY`

tanveergill (approved):
Please address comments and merge

pkg/policies/controlplane/validator_test.go
krdln: good catch!

krdln:
@sbienkow-ninja 
> I understand that the values that are read are coerced/deserialized to the target type regardless of the input type?

> JSON value will be a number or one of the special string values "NaN", "Infinity", and "-Infinity". **Either numbers or strings are accepted**. Exponent notation is also accepted. -0 is considered equivalent to 0. [[src](https://developers.google.com/protocol-buffers/docs/proto3#json)]

I guess using strings is beneficial, as they don't loose precision if passed through various implementations of json/yaml.

PR #364, opened 2022-09-07
Added validator and defaulter hooks for Agent and Controller CR

hdkshingala:
The PR contains below changes:
- Added Mutating webhook for Agent and Controller to set all the default values in their Custom Resource. This also does the validations so we can fail the CR before it gets created.
- Added pre-commit check for operator spec to verify that the CRD is not affected by any other changes.
- Added Image repository defaults for Agent and Controller CR
- Fixed the ConfigMaps unit tests

This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/364)


kklimonda-fn (approved):
Commit message is pretty bad - now that the repository is public and we are potentially going to be doing GitHub releases for operator, it would be good to make sure that there is a short explanation of the change, and not typos.

> Reviewed 63 of 63 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/364#-:-NBS0mWm-XaN02zlaBTj:b-2k0to1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hdkshingala)

harjotgill (approved):

PR #368, opened 2022-09-08
Add new simplified architecture diagram

krdln:
- rename old architecture diagrams
- add initial version from Jai

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/368)


harjotgill:
Worth showing etcd. i.e. etcd and prometheus are both user visible and are a dependency to run Aperture system...

harjotgill (approved):

krdln:
@harjotgill Oh, ok. So this time we're not doing "not naming specific technologies". I'll add ETCD tomorrow, although it will make it less obvious how controller talks to agent (unless that's not the point)

harjotgill:
We can name specific entries for the docs. Patent is a different story. :)

PR #374, opened 2022-09-08
Corrected some of the links in the docs

hdkshingala:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/374)


docs/content/get-started/installation/agent/agent.md
```diff
@@ -23,10 +23,10 @@ One Agent is deployed per node.
...
 
 The Aperture Agent related configurations are stored in a configmap which is
-created during the installation using Helm. All the configuration parameters are
-listed on the
-[README](https://artifacthub.io/packages/helm/aperture/aperture-operator#aperture-custom-resource-parameters)
-file of the Helm chart.
+created during the installation using Helm.
+
+All the configuration parameters for Aperture Agent are listed
+[here](reference/configuration/agent.md).
```
krdln, line 29: Hm, I thought absolute links need `/` at the beginning (eg. `[here](/reference/conf....md). I couldn't find this omitting `/` documented here https://docusaurus.io/docs/markdown-features/links, so perhaps let's stay safe and always start absolute paths with `/`?

krdln (approved):
Pushed one more fix for proto→reference link :)

You can fix these / or not – up to you.

docs/content/get-started/installation/agent/agent.md
hdkshingala: I think docusaurus checks it in the current directory and from root as well but yeah for safer side, added /.

PR #378, opened 2022-09-08
Move protobuf publishing to opsninja

kklimonda-fn:
There are multiple edge cases around publishing protobuf files to
buf.build:
- we want to tag buf.build registry with our release tag even if there
  were no changes to protobuf files.
- our final release will never have any changes from its matching
  release candidate.

Getting it right in bash would be a hassle, so make use of our
opsninja library to encapsulate entire logic.


sbienkow-ninja (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/378)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kklimonda-fn)

PR #385, opened 2022-09-08
extract compiler to a separate package for classifiers

harjotgill:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/385)


krdln (approved):
TBH this looks like a workaround, not a fix, but hey, it works :shrug:

Issue #447, opened 2022-09-13
Fix datarace in testing


harjotgill:
**Describe the bug**
```
=== RUN   TestOtel
Running Suite: Otel Suite - /root/project/plugins/service/aperture-plugin-fluxninja/otel
========================================================================================
Random Seed: �[1m1663088895�[0m

Will run �[1m5�[0m of �[1m5�[0m specs
{"level":"error","service":"otel-test","error":"open /etc/aperture/otel-test/config/otel-test.yaml: no such file or directory","file":"/etc/aperture/otel-test/config/otel-test.yaml","time":"2022-09-13T17:08:15Z","caller":"/root/project/pkg/filesystem/crud.go:52","message":"Failed to read from file"}
{"level":"error","service":"otel-test","error":"open /etc/aperture/otel-test/config/otel-test.yaml: no such file or directory","file":"/etc/aperture/otel-test/config/otel-test.yaml","time":"2022-09-13T17:08:15Z","caller":"/root/project/pkg/filesystem/crud.go:52","message":"Failed to read from file"}
{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
�[38;5;10m•�[0m{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
�[38;5;10m•�[0m{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
�[38;5;10m•�[0m{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
�[38;5;10m•�[0m{"level":"info","time":"2022-09-13T17:08:16Z","message":"Closing DiodeWriter after a delay!"}
==================
WARNING: DATA RACE
Write at 0x00c000b89270 by goroutine 418:
  runtime.closechan()
      /usr/local/go/src/runtime/chan.go:357 +0x0
  github.com/elastic/gmux.(*chanListener).Close.func1()
[...]

PR #450, opened 2022-09-14
Do not bring up platform in plugin UT

kwapik:
Bringing up platform in plugin UT is not necessary and can lead to data race when tests are ran in parallel.

Closes: GH-447

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/450)

### DIFF
--- a/plugins/service/aperture-plugin-fluxninja/otel/otel_test.go
+++ b/plugins/service/aperture-plugin-fluxninja/otel/otel_test.go
@@ -2,6 +2,7 @@ package otel_test
 
 import (
 	"context"
+	"encoding/json"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


krdln (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/450)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @krdln and @kwapik)

___
*[`plugins/service/aperture-plugin-fluxninja/otel/otel_test.go` line 52 at r1](https://reviewable.io/reviews/fluxninja/aperture/450#-NBvMK6Y17h21zE5Kepc:-NBvMK6Y17h21zE5Kepd:b-f83niu) ([raw file](https://github.com/fluxninja/aperture/blob/e66257dd5440cd36c338e7602e759e9209ceadf0/plugins/service/aperture-plugin-fluxninja/otel/otel_test.go#L52)):*
> ```Go
> 			func() config.Unmarshaller {
> 				return unmarshaller
> 			},
> ```

Nit: you should be able to put unmarshaller directy here using `fx.Supply`

PR #481, opened 2022-09-16
Split scraping metrics for FN plugin

kwapik:
##### Checklist

- [x] Breaking changes

### Description of change
Metrics scraped by prometheus have sampling rate determined by the scrape interval.
We don't want to send high sampled metrics to the FN Cloud, so in the plugin now duplicates the scraping config and changes the scrape interval.

This PR also changes the base scrape config to use only globally set scrape interval, instead of specyfing it per scrape target.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/481)

### DIFF
--- a/pkg/otel/config.go
+++ b/pkg/otel/config.go
@@ -248,7 +248,7 @@ func addPrometheusReceiver(cfg *otelParams) {
 	config.AddReceiver(ReceiverPrometheus, map[string]any{
 		"config": map[string]any{
 			"global": map[string]any{
-				"scrape_interval":     "1m",
+				"scrape_interval":     "1s",
 				"scrape_timeout":      "10s",
 				"evaluation_interval": "1m",
 			},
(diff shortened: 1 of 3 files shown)


pkg/otel/config.go
```diff
@@ -320,11 +319,10 @@ func buildApertureSelfScrapeConfig(name string, cfg *otelParams) map[string]any
...
 	return map[string]any{
-		"job_name":        "kubernetes-nodes",
-		"scheme":          "https",
-		"scrape_interval": "2s",
-		"scrape_timeout":  "1500ms",
-		"metrics_path":    "/metrics/cadvisor",
+		"job_name":       "kubernetes-nodes",
+		"scheme":         "https",
+		"scrape_timeout": "1500ms",
+		"metrics_path":   "/metrics/cadvisor",
```
tanveergill, line 325: How often are cadvisor metrics being scraped?

tanveergill (approved):

pkg/otel/config.go
kwapik: Every 1 second for control loop and every 10 seconds for FN plugin data.

PR #502, opened 2022-09-17
Refactor several APIs and structs

harjotgill:
##### Checklist

- [X] Tested in playground or other setup

### Description of change
Deep copy single entity vs. entire cache

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/502)


jaidesai-fn (approved):

api/aperture/common/peers/v1/peers.proto
```diff
@@ -3,14 +3,21 @@ syntax = "proto3";
...
 // PeerDiscoveryService is used to query Peers.
 service PeerDiscoveryService {
-  rpc GetPeers(PeersRequest) returns (Peers) {
+  rpc GetPeers(google.protobuf.Empty) returns (Peers) {
+    option (google.api.http) = {
+      get: "/v1/peers"
+    };
+  }
+
+  rpc GetPeer(PeerRequest) returns (Peer) {
```
seunghyupoh3517, line 20: Can the same API use to get both all the peers and specific peer was the question we had here - I ended up having repeated address field for the PeerRequest, is there different approach we can try?

harjotgill: repeated address field is not a good fit here. We need to get a single peer at a time and the peer database is not a nested map unlike status registry. e.g. for status we have nested keys (hence the repeated field) -- /v1/status/liveness/watchdog/heap and so on.

tanveergill (approved):

PR #547, opened 2022-09-20
add missing git commit hash and git branch in the script

slayer321:
##### Checklist

- [x] Tests and/or benchmarks are included

### Description of change

Add `GIT_COMMIT_HASH` and `GIT_BRANCH` in the script as earlier it was passed as an argument in the Dockerfile which was not given while running Tiltfile. Now this changes will directly take all the required values from the script and no need to pass it as the argument.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/547)

PR #549, opened 2022-09-20
add check for the git repo in the build script

slayer321:
##### Checklist

### Description of change

Add check for the git repo in the build script . As the recent changes made in #547 is failing CI in the Cloud repo. So added a check to only run the git command if we have the `.git` dir or else it will take is as the ARG  and also set it as ENV as previously in Cloud repo. 
cc @kwapik , @hdkshingala

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/549)

### DIFF
--- a/pkg/info/build.sh
+++ b/pkg/info/build.sh
@@ -10,8 +10,11 @@ GOARCH=$(go env GOARCH)
 HOSTNAME=$(hostname)
 SERVICE_FILE=$(basename -- "${TARGET}")
 SERVICE="${SERVICE_FILE%.*}"
-GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
-GIT_COMMIT_HASH=$(git rev-parse HEAD)
+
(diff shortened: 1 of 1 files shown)


pkg/info/build.sh
```diff
@@ -10,8 +10,11 @@ GOARCH=$(go env GOARCH)
 HOSTNAME=$(hostname)
 SERVICE_FILE=$(basename -- "${TARGET}")
 SERVICE="${SERVICE_FILE%.*}"
-GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
-GIT_COMMIT_HASH=$(git rev-parse HEAD)
+
+if [ -d .git ]; then
```
jmichalak-fluxninja, line 14: You can use `git rev-parse --is-inside-work-tree` for better robustness

jmichalak-fluxninja (approved):

PR #551, opened 2022-09-20
Added Init containers for Etcd and Prometheus in controller

hdkshingala:
##### Checklist

### Description of change

Added Init containers to check the health of embedded Etcd and Prometheus before starting the Controller container.to prevent unwanted controller pod restarts.

Fixes https://github.com/fluxninja/cloud/issues/5304

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/551)

### DIFF
--- a/manifests/charts/aperture-controller/templates/controller.yaml
+++ b/manifests/charts/aperture-controller/templates/controller.yaml
@@ -123,8 +123,41 @@ spec:
   {{- if .Values.controller.sidecars }}
   sidecars: {{ .Values.controller.sidecars | toYaml | nindent 4 }}
   {{- end }}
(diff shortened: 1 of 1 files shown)


harjotgill (commented):
Not sure if we need this --- what's wrong with restarts?

hdkshingala:
@harjotgill Usually when the Pod restarts, the first impression is that something is wrong with the application and it Crashed.

When the user installs the controller using the chart, the Etcd takes some time to start but our Controller is starting quickly and due to that I have seen the Controller Pod failing 3-4 times and than able to connect to Etcd and work properly.

This will prevent it so that the controller only fails when there is a genuine config error.

harjotgill (approved):

PR #621, opened 2022-09-29
use tag version for the helm chart release

slayer321:
##### Checklist

### Description of change

Earlier for the Helm chart release the version was taken from Chart.yaml . Which was not the effective as the version need to be change for every `rc` release in Chart.yaml. 
Now we are directly using the tag to create `.tgz`  package . Now we only need to change the version for the major release in Chart.yaml file .

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/621)


.circleci/scripts/helm_release.sh
```diff
@@ -51,8 +51,8 @@ dependencies() {
 
 package() {
   for chart in "${CHARTS[@]}"; do
-    CHART_VERSION_CMD=$(helm inspect chart "$chart" | grep version | cut -d' ' -f2|tr -d " \t\n\r")
-    APP_VERSION_CMD=$(helm inspect chart "$chart" | grep appVersion | cut -d' ' -f2|tr -d " \t\n\r")
+    CHART_VERSION_CMD=$(echo "$CIRCLE_TAG" | cut -d "/" -f 4)
+    APP_VERSION_CMD=$(echo "$CIRCLE_TAG" | cut -d "/" -f 4)
```
hdkshingala, line 55: Please add a check that the version should start with a `v` to make sure that it fails the script when the tag format is changed or incorrect version is provided.

kklimonda-fn (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/621)* status: all files reviewed, 1 unresolved discussion (waiting on @slayer321)

PR #663, opened 2022-10-04
add circleci pre-commit v2

slayer321:
##### Checklist

### Description of change

Fix all the minor issue that @hdkshingala mention . 
Earlier It was checking all the files inside `.circleci` dir.. Now it only checks the `*.*ml$` files . 
Add circleci cli install using asdf.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/663)

### DIFF
--- a/scripts/manage_tools.sh
+++ b/scripts/manage_tools.sh
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+declare -A plugin_repo_list
+plugin_repo_list[circleci_plugin]=https://github.com/lukeab/asdf-circleci-cli.git
+
 git_root="$(git rev-parse --show-toplevel)"
 readonly git_root="${git_root}"
 readonly version_file="${git_root}"/.tool-versions
@@ -43,7 +46,11 @@ add_plugins() {
 	for plugin in "${wanted_plugins[@]}"; do
 		if ! array_contains "${plugin}" "${added_plugins[@]}"; then
 			printf 'Adding asdf plugin: "%s"\n' "${plugin}"
-			asdf plugin add "${plugin}"
+			if [[ $plugin == "circleci" ]]; then
+				asdf plugin add circleci ${plugin_repo_list[circleci_plugin]}
+			else
+				asdf plugin add "${plugin}"
+			fi
 		fi
 	done
 }
--- a/.circleci/continue-workflows.yml
+++ b/.circleci/continue-workflows.yml
@@ -263,6 +263,7 @@ jobs:
             helm
             kustomize
             jsonnet
+            circleci
       - run:
           name: Install additional system dependencies for pre-commit
           command: |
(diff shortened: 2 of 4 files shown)


scripts/manage_tools.sh
```diff
@@ -43,7 +43,11 @@ add_plugins() {
 	for plugin in "${wanted_plugins[@]}"; do
 		if ! array_contains "${plugin}" "${added_plugins[@]}"; then
 			printf 'Adding asdf plugin: "%s"\n' "${plugin}"
-			asdf plugin add "${plugin}"
+			if [[ $plugin == "circleci" ]]; then
```
kklimonda-fn, line 46: Why are we we installing the plugin differently here?

kklimonda-fn: I think we should make this as a mapping somewhere, instead of doing if else dance in the function - either create a hashmap in manage_tools.sh, or add repository needed to install given to `.tool-versions` if possible.

@sbienkow-ninja any idea how we could more easily support asdf tools that require installation of plugins from specific repositories?

sbienkow-ninja (commented):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/663)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @kklimonda-fn and @slayer321)

___
*[`scripts/manage_tools.sh` line 46 at r2](https://reviewable.io/reviews/fluxninja/aperture/663#-NDXktYWEAvnlVuNTUJi-r2-46:-NE_O03o5i-VEY0vdks-:bgwhit6) ([raw file](https://github.com/fluxninja/aperture/blob/6f99c511a868060563d1c23640baa7ddae0117cf/scripts/manage_tools.sh#L46)):*
<details><summary><i>Previously, kklimonda-fn (Krzysztof Klimonda) wrote…</i></summary><blockquote>

I think we should make this as a mapping somewhere, instead of doing if else dance in the function - either create a hashmap in manage_tools.sh, or add repository needed to install given to `.tool-versions` if possible.

@sbienkow-ninja any idea how we could more easily support asdf tools that require installation of plugins from specific repositories?
</blockquote></details>

We could use an associative array to store such data (or even use a json file with jq), but I think this is fine for now?

slayer321:
yeah, we can use associative array to store the plugins links...
@sbienkow-ninja , @kklimonda-fn  I have made the suggested changes .. can you please review It .

kklimonda-fn (commented):
> Reviewed 3 of 4 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/663)* status: all files reviewed, 1 unresolved discussion (waiting on @slayer321)

kklimonda-fn (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/663)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @slayer321)

hdkshingala (approved):

PR #716, opened 2022-10-07
Add signals grafana dashboard

kklimonda-fn:
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/716)


harjotgill:
Would be good to add this to the playground as well...

harjotgill (approved):

PR #769, opened 2022-10-13
Switched to SSH instead of HTTP for GitHub operations

### LINKED ISSUE #768: Use ssh instead of http for GitHub connection in helm release process
Currently, we are using http for communicating with GitHub for releasing the charts to the `gh-pages` branch.
For this, we need to maintain a GitHib token which is not correct way to do it.

We need to replace it with ssh for which we already have keys stored in the CircleCi pipelines

hdkshingala:
### Description of change

We were using HTTP for communicating with GitHub which is incorrect as it requires providing the token which expires after some time.
Switched to using the SSH as we already store the SSH Keys for Aperture in CircleCi

Fixes #768

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/769)

### DIFF
--- a/.circleci/scripts/helm_release.sh
+++ b/.circleci/scripts/helm_release.sh
@@ -4,12 +4,12 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-REPO_URL="https://x-access-token:${GITHUB_TOKEN}@github.com/fluxninja/aperture.git"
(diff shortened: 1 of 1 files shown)


kklimonda-fn (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/769)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @hdkshingala)

PR #772, opened 2022-10-13
Readd Distcache metrics

seunghyupoh3517:
##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/772)

### DIFF
--- a/pkg/distcache/provider.go
+++ b/pkg/distcache/provider.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"strconv"
 	"sync"
-	"time"
 
 	"github.com/buraksezer/olric"
 	olricconfig "github.com/buraksezer/olric/config"
@@ -63,10 +62,9 @@ type DistCacheConfig struct {
 // DistCache is a peer to peer distributed cache.
 type DistCache struct {
 	sync.Mutex
-	Config   *olricconfig.Config
-	Olric    *olric.Olric
-	Metrics  *DistCacheMetrics
-	jobGroup *jobs.JobGroup
+	Config  *olricconfig.Config
+	Olric   *olric.Olric
+	Metrics *DistCacheMetrics
 }
 
 // AddDMapCustomConfig adds a named DMap config into DistCache's config.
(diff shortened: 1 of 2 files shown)


pkg/distcache/provider.go
```diff
@@ -60,8 +63,10 @@ type DistCacheConfig struct {
 // DistCache is a peer to peer distributed cache.
 type DistCache struct {
 	sync.Mutex
-	Config *olricconfig.Config
-	Olric  *olric.Olric
+	Config   *olricconfig.Config
+	Olric    *olric.Olric
+	Metrics  *DistCacheMetrics
+	jobGroup *jobs.JobGroup
```
harjotgill, line 69: not needed

harjotgill: use the existing liveness.service multijob instead

pkg/distcache/provider.go
```diff
@@ -173,27 +236,69 @@ func (constructor DistCacheConstructor) ProvideDistCache(in DistCacheConstructor
 
 	dc.Olric = o
 
+	dc.jobGroup = in.JobGroup
```
harjotgill, line 239: please don't add a new job group

pkg/distcache/provider.go
```diff
@@ -173,27 +236,69 @@ func (constructor DistCacheConstructor) ProvideDistCache(in DistCacheConstructor
...
+					log.Error().Err(err).Msg("Failed to start distcache")
+				}
 				_ = in.Shutdowner.Shutdown()
 			})
 			// wait for olric to start by waiting on startChan until ctx is canceled
 			select {
 			case <-ctx.Done():
 				return errors.New("olric failed to start")
 			case <-startChan:
+				err := dc.jobGroup.RegisterJob(job, jobConfig)
```
harjotgill, line 275: this can be moved outside the select statement. the select here is simply a barrier that waits for olric to start.

pkg/distcache/provider.go
```diff
@@ -173,27 +236,69 @@ func (constructor DistCacheConstructor) ProvideDistCache(in DistCacheConstructor
...
+					log.Error().Err(err).Msg("Failed to register distcache scrape metrics job with jobGroup")
+					return err
+				}
 			}
 			return nil
 		},
 		OnStop: func(ctx context.Context) error {
-			var multiErr error
+			// Unregister metrics with Prometheus.
+			err := dc.Metrics.unregisterMetrics(in.PrometheusRegistry)
```
harjotgill, line 285: these operations should be in reverse order of how they start - which means this should be called the last.

harjotgill (changes requested):

pkg/distcache/provider.go
```diff
@@ -172,11 +233,28 @@ func (constructor DistCacheConstructor) ProvideDistCache(in DistCacheConstructor
...
 		OnStart: func(ctx context.Context) error {
+			// Register metrics with Prometheus.
+			err := dc.Metrics.registerMetrics(in.PrometheusRegistry)
+			if err != nil {
+				log.Error().Err(err).Msg("Failed to register distcache metrics with Prometheus registry")
+				return err
+			}
+
 			panichandler.Go(func() {
 				err = dc.Olric.Start()
```
harjotgill, line 254: would be good to use a local variable for error as it might interfere with variable outside this go-routine.

seunghyupoh3517: I had to change the error name, `startErr`; otherwise, golangci yells due to shadow declaration

harjotgill (approved):

PR #906, opened 2022-11-08
Remove unused CheckResponse.Error

### LINKED ISSUE #894: Stop emitting error and non-error response simultaneusly in envoy authz Check
Eg. 
https://github.com/fluxninja/aperture/blob/a6424948f212ffb1aab7a88d94c7597595f8b4b4/pkg/policies/flowcontrol/api/envoy/authz.go#L168-L173

Grpc server _ignores_ the response if the error is not nil, so this response is not used in any way. We want to track the exact error reason though, preferably through metrics / logs (see also #882).

Would also be nice if returned errors wouldn't be just strings, but something that works with https://pkg.go.dev/google.golang.org/grpc/status#FromError, so that grpc server can send a meaningful response code.

krdln:
This field described only authz-specific errors and was filled in
envoy.Handler.Check() response when also returning non-nil error, but in
such case the grpc framework was not using the response anyway.
This field was also used for metrics, but no codepath was actually
setting them, as flowcontrol never set these.

Also:
* Create errors using grpc/status package, so that we have control on the grpc
  status.
* Add missing sampled logs for error conditions.

Drive-by:
* Remove unused error from ClassifierEngine.Classify(), as it's
  infallible (all errors are reported individually per-label).
* Remove unused code from authz.go.

Resolves #894

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/906)


kwapik (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/906)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @krdln)

___
*[`api/aperture/flowcontrol/check/v1/check.proto` line 39 at r1](https://reviewable.io/reviews/fluxninja/aperture/906#-NGNEOtM4UlBOgty1Pu9:-NGNNi7k7hfH4YW15mRQ:bxpnky0) ([raw file](https://github.com/fluxninja/aperture/blob/f4b603f6f04349b11594fc269a7b8c8659f65664/api/aperture/flowcontrol/check/v1/check.proto#L39)):*
<details><summary><i>Previously, krdln (Michał Krasnoborski) wrote…</i></summary><blockquote>

Don't see any compelling reason on why to renumber. Even though we don't care too much about compatibility, it won't hurt being more compatible than less :shrug: 

We could reserve, but since we're still alpha, I don't see too much value in it.
</blockquote></details>

I don't have strong opinion on that, so let's just keep it as is.

Issue #891, opened 2022-11-04
Add Armeria server example to the playground


krdln:
**Is your feature request related to a problem? Please describe.**
Right now tinkering with armeria integration requires manual deployment or running server locally with `kubectl port forward` (along with pre-defining a service). This adds friction to experimenting with armeria integration, especially to those doing it for the first time.

**Describe the solution you'd like**
Component on playground, with a button to trigger, similarly to how App.java is exposed as java sdk example. The component might be off by default.

cc @kwapik, @IridiumOxide

krdln:
Looks like @hasit already changed the main class from App to ArmeriaServer in https://github.com/fluxninja/aperture/pull/797/files. Clicking on the button makes the curl to decorated endpoint, but log exporting fails, probably due to misconfigured agent address:

```
[armeria-common-worker-epoll-2-1] INFO com.linecorp.armeria.internal.common.JavaVersionSpecific - Using the APIs optimized for: Java 12+
Nov 09, 2022 12:32:17 PM io.opentelemetry.sdk.internal.ThrottlingLogger doLog
SEVERE: Failed to export spans. Server is UNAVAILABLE. Make sure your collector is running and reachable from this network. Full error message:io exception
Running cmd: sh -c "\n# get k8s pod name from tilt resource name\nPOD_NAME=\"$(tilt get kubernetesdiscovery \"aperture-java-example\" -ojsonpath='{.status.pods[0].name}')\"\nkubectl exec -n aperture-java-example \"$POD_NAME\" -- wget http://127.0.0.1:8080/super -O -\n"
Connecting to 127.0.0.1:8080 (127.0.0.1:8080)
writing to stdout
-                    100% |********************************|    13  0:00:00 ETA
written to stdout
Hello, world!
Nov 09, 2022 12:32:24 PM io.opentelemetry.sdk.internal.ThrottlingLogger doLog
SEVERE: Failed to export spans. Server is UNAVAILABLE. Make sure your collector is running and reachable from this network. Full error message:io exception
```

Also, perhaps we want to have both "App" example and "ArmeriaServer" example as buttons?

hasit:
@krdln @slayer321 

Here's what we can do:
1. In `ArmeriaServer` example, add support for reading host and port of Aperture Agent to connect to via environment. See - https://github.com/fluxninja/aperture/blob/main/sdks/aperture-java/src/main/java/com/fluxninja/aperture/example/App.java#L18.
[...]

PR #913, opened 2022-11-10
Add ENV variable for armeria client and server

slayer321:
### Description of change
Fix https://github.com/fluxninja/aperture/issues/891

This PR fix the first part of the issue mention by @hasit  . Where ENV variable where missing from armeria client and server due to which aperture-java sdk was failing on Tilt playground when user clicked on `trigger library example`

as discuss with @krdln , raising this PR... second part will be done by @IridiumOxide soon.

cc @krdln 

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/913)


krdln (approved):
Cool! Verified that fluxmeters etc work on playground when using `service: aperture-java-example.aperture-java-example.svc.cluster.local` selector

Issue #883, opened 2022-11-02
Warn on incompatible k8s version


krdln:
### Is your feature request related to a problem? Please describe.
Supported k8s versions are clearly documented, but it's easy to accidently deploy on old version and the failure mode is hard to debug (see #800).

### Describe the solution you'd like
In-your-face warning about using old version of k8s. Not sure what's the most appropriate place would be. My suggestions:
* warn on agent startup,
* warn on helm deployment,
* UI?

chiukapoor:
Hello @krdln, For the `warn on agent startup` I am working on creating an event to be generated by the Agent and Controller if in case the minimum Kubernetes version is not satisfied.

To grab the current Kubernetes version will be following the same method as `kubectl`

https://github.com/kubernetes/kubectl/blob/393c40f5c4acbe48edbc70f8c8696bb623744b76/pkg/cmd/version/version.go#L139

chiukapoor:
Agent aperture control manager log:

```
1.6685863398395433e+09	DEBUG	events	Kubernetes version v1.25.0 is not supported. Please use a kubernetes cluster with version v1.27.0 or above	{"type": "Warning", "object": {"kind":"Agent","namespace":"aperture-agent","name":"agent","uid":"ffe0db7d-fcf5-4685-84b2-da04ae7534f6","apiVersion":"fluxninja.com/v1alpha1","resourceVersion":"14668896"}, "reason": "MinimumKubernetesVersionFail"}
```

controller-aperture-contoller-manager

```
1.6685863398325784e+09	DEBUG	events	Kubernetes version v1.25.0 is not supported. Please use a kubernetes cluster with version v1.27.0 or above	{"type": "Warning", "object": {"kind":"Controller","namespace":"aperture-controller","name":"controller","uid":"26df4d5c-56af-453e-be98-68a1e1ecf63b","apiVersion":"fluxninja.com/v1alpha1","resourceVersion":"14668895"}, "reason": "MinimumKubernetesVersionFail"}
```

PR #933, opened 2022-11-16
Added the functionality to check for the minimum supported kubernetes version

chiukapoor:
Signed-off-by: Chirayu Kapoor <dev.csociety@gmail.com>

### Description of change

Fixes: https://github.com/fluxninja/aperture/issues/883

Created an event to be generated by the Agent and Controller if in case the minimum Kubernetes version is not satisfied.

To grab the current Kubernetes version followed the same method as kubectl: https://github.com/kubernetes/kubectl/blob/393c40f5c4acbe48edbc70f8c8696bb623744b76/pkg/cmd/version/version.go#L139

aperture-agent-manager log:

```
1.6685863398395433e+09	DEBUG	events	Kubernetes version v1.25.0 is not supported. Please use a kubernetes cluster with version v1.27.0 or above	{"type": "Warning", "object": {"kind":"Agent","namespace":"aperture-agent","name":"agent","uid":"ffe0db7d-fcf5-4685-84b2-da04ae7534f6","apiVersion":"fluxninja.com/v1alpha1","resourceVersion":"14668896"}, "reason": "MinimumKubernetesVersionFail"}
```

aperture-controller-manager log:

```
1.6685863398325784e+09	DEBUG	events	Kubernetes version v1.25.0 is not supported. Please use a kubernetes cluster with version v1.27.0 or above	{"type": "Warning", "object": {"kind":"Controller","namespace":"aperture-controller","name":"controller","uid":"26df4d5c-56af-453e-be98-68a1e1ecf63b","apiVersion":"fluxninja.com/v1alpha1","resourceVersion":"14668895"}, "reason": "MinimumKubernetesVersionFail"}
```

The minimum version was kept `1.27.0` for testing purposes

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933)

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -2,3 +2,4 @@ inport
 outport
 flowcontrolv1
 Fatalf
+AtLeast
--- a/.codespellignore
+++ b/.codespellignore
@@ -2,4 +2,4 @@ inport
 outport
 flowcontrolv1
 Fatalf
-AtLeast
+atleast
--- a/operator/controllers/agent/reconciler.go
+++ b/operator/controllers/agent/reconciler.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/strings/slices"
+	apimachineryversion "k8s.io/apimachinery/pkg/version"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -166,6 +167,12 @@ func (r *AgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		} else if instance.Status.Resources == controllers.FailedStatus {
 			return ctrl.Result{}, nil
 		}
+	}
+
+	// Checking if the Minimum kubernetes version condition is satisfied.
+	if len(instance.Status.Resources) == 0 && apimachineryversion.CompareKubeAwareVersionStrings(controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion) == 1 {
+		r.Recorder.Event(instance, corev1.EventTypeWarning, "MinimumKubernetesVersionFail",
+		fmt.Sprintf("Kubernetes version %v is not supported. Please use a kubernetes cluster with version %v or above", controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion))
 	}
 
 	instance.Status.Resources = "creating"
--- a/operator/controllers/agent/reconciler.go
+++ b/operator/controllers/agent/reconciler.go
@@ -32,7 +32,6 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	apimachineryversion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/strings/slices"
@@ -170,9 +169,9 @@ func (r *AgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	}
 
 	// Checking if the Minimum kubernetes version condition is satisfied.
-	if len(instance.Status.Resources) == 0 && apimachineryversion.CompareKubeAwareVersionStrings(controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion) == 1 {
+	if len(instance.Status.Resources) == 0 && !controllers.MinimumKubernetesVersionBool {
 		r.Recorder.Event(instance, corev1.EventTypeWarning, "MinimumKubernetesVersionFail",
-			fmt.Sprintf("Kubernetes version %v is not supported. Please use a kubernetes cluster with version %v or above", controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion))
+			fmt.Sprintf("Kubernetes version %v is not supported. Please use a kubernetes cluster with version %v or above.", controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion))
 	}
 
 	instance.Status.Resources = "creating"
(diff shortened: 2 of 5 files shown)


chiukapoor:
PTAL

@krdln

kwapik:
@chiukapoor something got messed up with the CI results, but please resolve issues from [here](https://app.circleci.com/pipelines/github/fluxninja/aperture/4571/workflows/bd2df552-45b7-46da-92e9-3613d6293425/jobs/16168).
To have those fixed in pre-commit hooks, follow https://github.com/fluxninja/aperture/blob/main/CONTRIBUTING.md

chiukapoor:
Hello @kwapik, the issue is fixed now.

chiukapoor:
My bad I didn't install the pre-commit hooks.

Thanks @kwapik and @hdkshingala for the help!

hdkshingala (approved):

hdkshingala:
LGTM.

@krdln Could you please check if this is what we wanted as a solution?

krdln (changes requested):
looks good! Although, if someone is not looking at the alerts, the debug level can easily be missed. Can we additionally log a _normal_ warning (`log.Warn()`)? Not sure how often the reconcile is running – we don't want to spam the logs.

> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933#-:-NH3rpXu8kgzfrhRdW14:b-d54g2)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @chiukapoor)

___
*[`operator/controllers/agent/reconciler.go` line 173 at r1](https://reviewable.io/reviews/fluxninja/aperture/933#-NH3r7uf32iiQfjrBCmE:-NH3r7uf32iiQfjrBCmF:b-dqdt5w) ([raw file](https://github.com/fluxninja/aperture/blob/10e05d91679d6138fa56c587cccfdb792db89f35/operator/controllers/agent/reconciler.go#L173)):*
> ```Go
> 
> 	// Checking if the Minimum kubernetes version condition is satisfied.
> 	if len(instance.Status.Resources) == 0 && apimachineryversion.CompareKubeAwareVersionStrings(controllers.CurrentKubernetesVersion.String(), controllers.MinimumKubernetesVersion) == 1 {
> ```

I think the comparison should be `>= 0`, not `== 1` (because `0` means equal, and that's also ok, and if version is greater, we might get other positive number, not only `1`).

hdkshingala:
@krdln Regarding the log, we can add a log line for warn which will be executed only once as the above if checks the status field which will always have some value after first reconciliation.

FYI, this log will be on the operator and we are adding an event on the CR so it'll be visible when someone does `kubectl describe`.

krdln (commented):
If you say the visibility is good enough, then we don't need additional logs. Please see comment about the comparison though.

> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933#-:-NH3uCS3CEuB9PwmWK4I:bdzmljr)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @chiukapoor)

chiukapoor:
> I think the comparison should be >= 0, not == 1 (because 0 means equal, and that's also ok, and if version is greater, we might get other positive number, not only 1).

@krdln The `apimachineryversion.CompareKubeAwareVersionStrings` function internally uses [strings.Compare](https://go.dev/src/strings/compare.go) for kubernetes version string and the `Compare` function only returns 0, 1 or -1, In which 1 stands for `string1>string2`

krdln (commented):
@chiukapoor 
I think it uses a bit more complex logic, see:

```go
// CompareKubeAwareVersionStrings compares two kube-like version strings.
// Kube-like version strings are starting with a v, followed by a major version, optional "alpha" or "beta" strings
// followed by a minor version (e.g. v1, v2beta1). Versions will be sorted based on GA/alpha/beta first and then major
// and minor versions. e.g. v2, v1, v1beta2, v1beta1, v1alpha1.
func CompareKubeAwareVersionStrings(v1, v2 string) int {
	if v1 == v2 {
		return 0
	}
	v1major, v1type, v1minor, ok1 := parseKubeVersion(v1)
	v2major, v2type, v2minor, ok2 := parseKubeVersion(v2)
	switch {
	case !ok1 && !ok2:
		return strings.Compare(v2, v1)
	case !ok1 && ok2:
		return -1
	case ok1 && !ok2:
		return 1
	}
	if v1type != v2type {
		return int(v1type) - int(v2type)
	}
	if v1major != v2major {
		return v1major - v2major
	}
	return v1minor - v2minor
}
```

Anyway, we still want to allow version == minimum, so you need to also support `CompareKubeAwareVersionStrings` returning `0`.

> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933#-:-NH440hk7iejbpGsgwNM:bqm7xe1)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @chiukapoor)

chiukapoor:
The function `parseKubeVersion` run the string through a regex which check if the string is of type "v1alpha1" or "v1beta1" etc. As our passed kubernetes version string does not match the regex it returns `0, 0, 0, false`. 

https://github.com/kubernetes/apimachinery/blob/4fbe8e4db39b17a53c0756caf58466baad871871/pkg/version/helpers.go#L36-L41

As both the `ok` are false, code below checks the strings using `string.Compare`

```
	v1major, v1type, v1minor, ok1 := parseKubeVersion(v1)
	v2major, v2type, v2minor, ok2 := parseKubeVersion(v2)
	switch {
	case !ok1 && !ok2:
		return strings.Compare(v2, v1)
```

@krdln, After thoroughly examining the `apimachineryversion.CompareKubeAwareVersionStrings` function, I believe we may instead use the `strings.Compare` function to perform the version check. Please let me know your thoughts on this.

> Anyway, we still want to allow version == minimum, so you need to also support CompareKubeAwareVersionStrings returning 0.

Will inculcate the suggestion

krdln (commented):
Yeah, looking at this `CompareKubeAwareVersionStrings`, looks like it's used to compare `v3alpha1` versions that are used for CRD versions, so yeah, I agree that this function shouldn't be used here at all.

`strings.Compare` will work in most cases, but eg. `"v1.9.0" > "v1.23.0"`.
Perhaps this `AtLeast` function will work? https://github.com/kubernetes/apimachinery/blob/master/pkg/util/version/version.go#L305

> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933#-:-NH4L3ffDfEu2neTmyx0:bhpp3qi)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @chiukapoor)

krdln (approved):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/933)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @chiukapoor)

krdln:
@chiukapoor To fix codespell pre-commit issue, you need to make `atleast` in .codespell ignore lowerase. Thanks!

Issue #608, opened 2022-09-27
Alerter Component


DariaKunoichi:
## Spec
* in_ports
  * alert
* alert_name (string): Name of the Alert
* severity (string): One of `info`, `warn` or `crit`
* resolve_timeout (duration): Defaults to `5m`
* alert_channels: A list of alert channel strings

## Runtime
* If alert signal is true, generate an Alert event, refer https://github.com/fluxninja/aperture/issues/861
  * Labels: `policy_name`, `type:alerter`, `alert_name`, `severity`
  * Additional labels added by FluxNinja plugin: `controller_id`
  * Annotations: `alert_channels`, `resolve_timeout`
* Else, do nothing

PR #939, opened 2022-11-16
Alerter Component

DariaKunoichi:
### Description of change
Closes: GH-608

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/939)


pkg/policies/controlplane/component-factory.go
```diff
@@ -161,6 +163,15 @@ func NewComponentAndOptions(
 			Name:          "Sink",
 			ComponentType: runtime.ComponentTypeSink,
 		}, nil, option, err
+	} else if alerter := componentProto.GetAlerter(); alerter != nil {
+		component, option, err := components.NewAlerterAndOptions(alerter, componentIndex, policyReadAPI, alerterIface)
```
tanveergill, line 167: Can alerter be provided via Fx just like etcd client, Prometheus registry etc. All other component constructors have a consistent and minimal signature so that we can introduce custom components in the future. Hope we can keep it that way

DariaKunoichi: @tanveergill fixed it

tanveergill (approved):

Issue #882, opened 2022-11-02
Sanity metrics


krdln:
### Is your feature request related to a problem? Please describe.
We need some sanity metrics to be able to quickly assess if agent's data processing pipeline is behaving in expected way. Eg. such metrics would make debugging #800 a lot easier.

#### Following metrics are pushed to prometheus:
* "flowcontrol_requests_total" & "flowcontrol_decisions_total" – these are already available.
* \# of failed IP lookup on Authorization.Check
* \# of failed IP lookup on Flowcontrol.Check
* \# of failed IP lookup on logs processors
* \# of missing CheckResponse in logs processor
* \# of logs processed.

Alternatively – various errors of authz / flowcontrol / logs processing could be accumulated as a single metric with differing error attributes (similarly to flowcontrol_decisions_total). Unsure if that flexibility is necessary though – we need "failed IP lookup" and "missing CheckResponse" the most, as these errors basically break the pipeline. All the other errors are (?) non-pipeline-breaking and thus can be debugged eg. in druid.

#### These metrics should be visible on some pre-shipped dashboard:

Dashboard should show values accumulated from all the agents, but per-agent filtering should also be possible.

Error values should be visible on dashboard both as counts and percentages.

krdln:
I've expanded the scope of this issue to include various improvements of handling error conditions in datapath in general – as these are a soft prerequisite to having good metrics.

PR #943, opened 2022-11-17
Start warning on entity lookup failure in authz and flowcontrol

krdln:
### Description of change

Added warnings on client IP retrieval failure and entity lookup failure.
Also, converted these failures to fail-early, so they're more
dicsoverable. We might lose some things (like stats) by not processing
such request further, but since we don't have any metadata about
services, such information was not too helpful.

This is a step forward to #882.

Drive-by:
* Ripped out comments suggesting that on nil EntityCache we do some sort of fallback service detection via Host – this logic is long gone now.
* Entity cache is assumed to be non-nil.
* Added helpers in grpc package to simplify "log and return grpc error reusing the same message".
* Refactored if-else to switch-case in authz Check, as `DecisionType != REJECTED` was weird to read.
* Removed Eventually() in authz tests, it's no longer needed from some time.

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/943)

### DIFF
--- a/pkg/policies/flowcontrol/service/envoy/authz.go
+++ b/pkg/policies/flowcontrol/service/envoy/authz.go
@@ -14,10 +14,9 @@ import (
 	"github.com/open-policy-agent/opa/logging"
 	"google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/genproto/googleapis/rpc/status"
-	grpc_codes "google.golang.org/grpc/codes"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
-	grpc_status "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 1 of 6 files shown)


kwapik (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/943)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @krdln)

pkg/policies/flowcontrol/service/envoy/authz.go
```diff
@@ -139,48 +133,45 @@ func (h *Handler) Check(ctx context.Context, req *ext_authz.CheckRequest) (*ext_
...
-		}
+	if !peerExists {
+		return nil, grpc.Bug().Msg("cannot get peer info")
 	}
 
+	clientIP := strings.Split(rpcPeer.Addr.String(), ":")[0]
+	entity, err := h.entityCache.GetByIP(clientIP)
+	if err != nil {
+		return nil, grpc.LoggedError(log.Sample(unknownEntitySampler).Warn()).
+			Str("IP", clientIP).Code(codes.NotFound).Msg("unknown entity")
```
tanveergill, line 154: Entity lookup failure should be treated fatal. Service discovery is disabled in sidecar based installation. The catch all service rule still applies.

If there is a genuine discovery issue, it can be discovered via Flow Analytics or via other mechanisms

PR #946, opened 2022-11-18
`git push` retry for Helm-release script  

slayer321:
### Description of change
Currently when we release aperture, 3 pipelines make changes to same branch which leads to conflict and git push fails.
If  `git push` fails due to conflict the added check will retry it for 10 times.

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/946)

### DIFF
--- a/.circleci/scripts/helm_release.sh
+++ b/.circleci/scripts/helm_release.sh
@@ -19,7 +19,16 @@ main(){
   locate
   dependencies
   package
-  upload
+
+  retry_counter=10
+  while true;do
+    (( retry_counter-- )) || break
+    if upload; then
+      echo "git push passed"
+      break
+    fi
+    sleep 1
+  done
 }
 
 install_helm(){
@@ -74,6 +83,7 @@ package() {
 upload(){
   tmpDir=$(mktemp -d)
   pushd "$tmpDir" >& /dev/null
+  trap 'popd &>/dev/null && rm -rf "$tmpDir"' RETURN
 
   git clone "${REPO_URL}"
   cd aperture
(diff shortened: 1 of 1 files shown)


sbienkow-ninja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/946)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @slayer321)

___
*[`.circleci/scripts/helm_release.sh` line 26 at r1](https://reviewable.io/reviews/fluxninja/aperture/946#-NH9gZCw9y7PxI2tcXRM:-NH9gZCw9y7PxI2tcXRN:b3xwntn) ([raw file](https://github.com/fluxninja/aperture/blob/90904b8bf3dd0296c6a6ebc27e828f42f24ea0d3/.circleci/scripts/helm_release.sh#L26)):*
> ```Shell
>   while true;do
>     (( retry_counter-- )) || break
>     set +e
> ```

This `set +e` and `set -e` should not be here - they only allow for `upload` to silently continue on error - we should explicitly handle such cases.
___
*[`.circleci/scripts/helm_release.sh` line 120 at r1](https://reviewable.io/reviews/fluxninja/aperture/946#-NH9g-DPEO7AmdQNEkOk:-NH9g-DPEO7AmdQNEkOl:b-tvpbw6) ([raw file](https://github.com/fluxninja/aperture/blob/90904b8bf3dd0296c6a6ebc27e828f42f24ea0d3/.circleci/scripts/helm_release.sh#L120)):*
> ```Shell
>   fi
>   popd >& /dev/null
>   rm -rf "$tmpDir"
> ```

Now we are both changing state and leaving tmp trash - perhaps both `popd` and `rm` commands should be added to a `RETURN` trap?



_Code quote:_
```Shell
  if git push origin ${BRANCH}; then
    echo "git push passed"
  else
    return 1
  fi
  popd >& /dev/null
  rm -rf "$tmpDir"
```

slayer321:
Yeah, `pod` and `rm` should be run before return ..as it is getting created everytime we run the upload function .

sbienkow-ninja (changes requested):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/946)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @slayer321)

___
*[`.circleci/scripts/helm_release.sh` line 120 at r1](https://reviewable.io/reviews/fluxninja/aperture/946#-NH9g-DPEO7AmdQNEkOk:-NH9nm3p6wVDH6-7n2cn:bmnadkz) ([raw file](https://github.com/fluxninja/aperture/blob/90904b8bf3dd0296c6a6ebc27e828f42f24ea0d3/.circleci/scripts/helm_release.sh#L120)):*
<details><summary><i>Previously, sbienkow-ninja (Szymon Bieńkowski) wrote…</i></summary><blockquote>

Now we are both changing state and leaving tmp trash - perhaps both `popd` and `rm` commands should be added to a `RETURN` trap?
</blockquote></details>

@slayer321 now we don't clean up if the push passed :/
It might be worth adding something like this to the top of the function:
```
upload(){
  tmpDir=$(mktemp -d)
  pushd "$tmpDir" >& /dev/null
  trap 'popd &>/dev/null && rm -rf "$tmpDir"' RETURN
```
This will make sure it always runs, even if it fails earlier doing copying or whatever.

sbienkow-ninja (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/946)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @slayer321)

PR #961, opened 2022-11-23
Add an easier way of defining test scenarios for aperture

kklimonda-fn:
### Description of change

This is still WIP with one largest problem being no support for tanka - a bit tricky given how our tanka apps can have both jsonnet and chart dependencies, which live in a different place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/961)


tanveergill:
@kklimonda-fn,

You can close this PR if you don't think this approach is viable

Issue #905, opened 2022-11-08
Garbage collector in ETCD


kwapik:
**Is your feature request related to a problem? Please describe.**
Keys in ETCD are not cleaned up. This may lead to disk space running out quite quickly.

**Describe the solution you'd like**
Some kind of Garbage collector in Controller/Agent (IMO the one which is writing should clean up its part).

kklimonda-fn:
An alternative to doing compaction ourselves, would be to document required etcd autocompaction configuration.

tanveergill:
@kwapik,
Which keys are not getting cleaned up? The Policy decisions are written under a lease and should get auto-removed once the Policy is deleted unless we missed something

kwapik:
@kklimonda-fn in which env I can check that?

hdkshingala:
@kwapik It's on the customer demo cluster

hdkshingala:
The DB snapshot is uploaded at https://drive.google.com/file/d/1lg0DEMzIme0Vuie-gkrNsYslaTV58C4C/view?usp=share_link.

tanveergill:
That is correct. We only ever need the latest value of each key

kwapik:
@kklimonda-fn so we just need to enable compaction by default.

DariaKunoichi:
@tanveergill there is also an issue of configuring this in our etcd instance or in etcd client - what if customer wants to provide their etcd without compacting? We will then probably flood their etcd instance.

tanveergill:
> @tanveergill there is also an issue of configuring this in our etcd instance or in etcd client - what if customer wants to provide their etcd without compacting? We will then probably flood their etcd instance.

@DariaKunoichi,
I looked at the client side compaction API. It works only on the global revision number. If someone is using the same etcd db for multiple workload, we cannot invoke compaction for Aperture's keyspace only. We should let the user configure compaction on their own etcd based on all their workloads

PR #992, opened 2022-12-01
Garbage collector in ETCD

DariaKunoichi:
### Description of change
Closes: GH-905

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/992)

### DIFF
--- a/manifests/charts/aperture-controller/values.yaml
+++ b/manifests/charts/aperture-controller/values.yaml
@@ -542,6 +542,8 @@ controller:
 ##
 etcd:
   enabled: true
+  autoCompactionMode: periodic
+  autoCompactionRetention: 24
   initContainer:
     enabled: true
     image:


manifests/charts/aperture-controller/values.yaml
```diff
@@ -542,6 +542,8 @@ controller:
 ##
 etcd:
   enabled: true
+  autoCompactionMode: periodic
+  autoCompactionRetention: 10m
```
hdkshingala, line 546: The description for this in etcd chart says this value should be in hour:
```
@param autoCompactionRetention Auto compaction retention for mvcc key value store in hour,
```
So, I think we can set default to 24 hour as we have used the same in our customer demo cluster where we have the wavepool generator 24x7 and we haven't faced the size issue after setting that.

harjotgill (approved):

PR #1033, opened 2022-12-15
rollupprocessor: Prepare for multiple stages of rollup

krdln:
### Description of change

Stage field in rollupprocessor config decides whether processor should perform
initial rollup from logs, or "rollup of rollups". Each Rollup now "knows" how
it should be aggregated.

Aggregate rollup is not enabled in any pipeline yet (and can't, as we need to
implement RollupDatasketchMerge).

---

Wanted to represent rollup of counts as RollupSum, but RollupSum operates on floats. We can either:
* special-case counts, don't try to treat it as regular Rollup,
* duplicate all float-attribute-related code to work on int-fields too,
* just accept that sum of counts won't be int.

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1033)


tanveergill:
@krdln,
What’s the blocked on? Are you waiting for reviews?

krdln:
@tanveergill Wanted to get feedback from @kwapik if this is approach we want (the PR itself was done as a drive-by)

Issue #950, opened 2022-11-20
Differentiator Component 


tanveergill:
## Spec
* in_ports:
  * input: The input signal
* out_ports:
  * output
* window: The window of time over which differentiator operates

## Runtime
* Calculates rate of change per tick

PR #1058, opened 2022-12-28
Add Differentiator component

DariaKunoichi:
### Description of change
Closes: GH-950

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1058)


pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (d *Differentiator) extrapolate(firstVal, secondVal point, currentTime time.Time) float64 {
+	extValue := firstVal.reading.Value() + float64(currentTime.UnixMilli()-firstVal.timestamp.UnixMilli())/
+		float64(secondVal.timestamp.UnixMilli()-firstVal.timestamp.UnixMilli())*
+		(secondVal.reading.Value()-firstVal.reading.Value())
+	return extValue
+}
+
+func (d *Differentiator) init(tickInfo runtime.TickInfo) {
+	d.capacity = int(d.window / tickInfo.Interval())
```
tanveergill, line 115: Instead of rounding down, take the ceiling similar to how we do it elsewhere, such as [EMA window](https://github.com/fluxninja/aperture/blob/main/pkg/policies/controlplane/components/ema.go#L62)

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+	d.readings = make([]point, d.capacity)
+	for i := 0; i < d.capacity; i++ {
+		d.readings[i].reading = runtime.InvalidReading()
+		d.readings[i].timestamp = time.Now()
```
tanveergill, line 119: Why use wallclock time anywhere?

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+func (d *Differentiator) Execute(inPortReadings runtime.PortToValue, tickInfo runtime.TickInfo) (runtime.PortToValue, error) {
+	if !d.initialized {
+		d.init(tickInfo)
+	}
+
+	inputVal := inPortReadings.ReadSingleValuePort("input")
+	outputVal := runtime.NewReading(0)
+
+	// add input to readings array
+	d.readings[d.newestIdx] = point{reading: inputVal, timestamp: tickInfo.Timestamp()}
```
tanveergill, line 65: Maybe you forgot to initialize `newestIdx` and `oldestIdx` based on capacity? They would always be zero

Could have used a deque like structure for a simpler implementation?

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+			if found.reading.Valid() && oldest.reading.Valid() {
+				extrapolatedValue := d.extrapolate(oldest, found, tickInfo.Timestamp())
+				newest = point{reading: runtime.NewReading(extrapolatedValue), timestamp: tickInfo.Timestamp()}
+			}
+		}
+
+		// calculate the derivative
+		if oldest.reading.Valid() && newest.reading.Valid() {
+			diff := (newest.reading.Value() - oldest.reading.Value()) /
+				float64(newest.timestamp.UnixMilli()-oldest.timestamp.UnixMilli())
```
tanveergill, line 87: Not supposed to divide by time: ```Differentiator calculates rate of change per tick```

Simply divide by number of ticks (capacity)

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Execute implements runtime.Component.Execute.
+func (d *Differentiator) Execute(inPortReadings runtime.PortToValue, tickInfo runtime.TickInfo) (runtime.PortToValue, error) {
+	if !d.initialized {
+		d.init(tickInfo)
+	}
+
+	inputVal := inPortReadings.ReadSingleValuePort("input")
+	outputVal := runtime.NewReading(0)
```
tanveergill, line 62: Why zero? Should emit invalid if unable to compute diff

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// add input to readings array
+	d.readings[d.newestIdx] = point{reading: inputVal, timestamp: tickInfo.Timestamp()}
+
+	if d.oldestIdx != d.newestIdx {
```
tanveergill, line 67: What does this signify?

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+		newest := d.readings[d.newestIdx]
+
+		if !oldest.reading.Valid() {
+			oldest = d.firstValid(d.oldestIdx, true)
+		}
+
+		// find the newest valid reading for extrapolation
+		if !newest.reading.Valid() {
+			found := d.firstValid(d.newestIdx, false)
+			if found.reading.Valid() && oldest.reading.Valid() {
```
tanveergill, line 78: Cannot compute diff if oldest and newest first valid happens to be same id. Should emit invalid in that case

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+func (d *Differentiator) firstValid(fromIdx int, addition bool) point {
+	step := 1
+	if !addition {
+		step = -1
+	}
+
+	idx := (fromIdx + step) % d.capacity
+	if (fromIdx + step) < 0 {
+		// ring buffer wrap up
+		idx = d.capacity - 1
```
tanveergill, line 133: doesn't modulo already handle this fine?

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for idx != fromIdx {
+		if d.readings[idx].reading.Valid() {
+			return d.readings[idx]
+		}
+		idx = (idx + step) % d.capacity
```
tanveergill, line 140: Please eliminate repetition of this clause [above](https://github.com/fluxninja/aperture/pull/1058/files#diff-e29f5f51a11446300c7fffccc1c564f797396c4b1b24e07214eec5beeebc3a78R130) by running of loop of capacity instead of the current terminating condition

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+		if (idx + step) < 0 {
+			// ring buffer wrap up
+			idx = d.capacity - 1
+		}
```
tanveergill, line 144: Should not be required since module takes care of it

pkg/policies/controlplane/components/differentiator.go
```diff
@@ -0,0 +1,147 @@
...
+	oldestIdx int
+	newestIdx int
+	// capacity is calculated from window duration divided by tick interval
+	capacity    int
+	initialized bool
+}
+
+type point struct {
+	reading   runtime.Reading
+	timestamp time.Time
```
tanveergill, line 29: Don't need to track the timestamps

tanveergill (approved):
Okay to merge once comments are addressed

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: ok

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: okay, I though that for differentiator the default value would always be 0 == "no change"

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: ok

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: that it's not our very first reading

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: ok

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: unfortunately in golang modulo can produce negative values, it does not behave as I would expect

pkg/policies/controlplane/components/differentiator.go
DariaKunoichi: ok

pkg/policies/controlplane/components/differentiator.go
tanveergill: Can use [this](https://stackoverflow.com/a/59299881) as the Mod func:
```
func mod(a, b int) int {
    return (a % b + b) % b
}
```

tanveergill: Add it to utils

PR #1068, opened 2022-12-30
Fix providing control point cache

DariaKunoichi:
### Description of change
Fixed providing control point cache in agent and controller.

##### Checklist

### DIFF
--- a/cmd/aperture-controller/controller.go
+++ b/cmd/aperture-controller/controller.go
@@ -14,10 +14,12 @@ import (
 	"go.uber.org/fx"
 
 	"github.com/fluxninja/aperture/cmd/aperture-controller/controller"
+	"github.com/fluxninja/aperture/pkg/cache"
 	"github.com/fluxninja/aperture/pkg/log"
 	"github.com/fluxninja/aperture/pkg/otelcollector"
 	"github.com/fluxninja/aperture/pkg/platform"
 	"github.com/fluxninja/aperture/pkg/policies/controlplane"
+	"github.com/fluxninja/aperture/pkg/policies/flowcontrol/selectors"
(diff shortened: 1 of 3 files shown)


jmichalak-fluxninja (approved):

cmd/aperture-controller/controller.go
```diff
@@ -28,6 +30,7 @@ func main() {
 		controller.ModuleForControllerOTEL(),
 		fx.Provide(
 			clockwork.NewRealClock,
+			cache.Provide[selectors.ControlPointID],
```
tanveergill, line 33: Why is Control Point cache required in the Controller?

DariaKunoichi: It's needed if we enable plugins

DariaKunoichi: You changed the ControlPointCache in Heartbeats to the new one, but did not provide it for controller.

tanveergill: But Controller does not do Service Control Point discovery.

Perhaps we ca mark ControlPointCache as optional [here](https://github.com/fluxninja/aperture/blob/main/plugins/service/aperture-plugin-fluxninja/heartbeats/provide.go#L57) and do a nil check in Heartbeat [here](https://github.com/fluxninja/aperture/blob/main/plugins/service/aperture-plugin-fluxninja/heartbeats/heartbeats.go#L225)?

DariaKunoichi: Sure, we can, can you create an issue to track it?

tanveergill: https://github.com/fluxninja/aperture/issues/1069

PR #1105, opened 2023-01-10
Replace Optimizer sub circuit logic with Integrator component

DariaKunoichi:
### Description of change
Closes: GH-1059

##### Checklist

- [ ] Breaking changes


blueprints/lib/1.0/blueprints/latency-gradient/policy.libsonnet
```diff
@@ -55,18 +56,15 @@ function(params) {
...
-          ),
-          component.withFirstValid(
-            firstValid.new()
-            + firstValid.withInPorts({ inputs: [port.withSignalName('CONCURRENCY_INCREMENT_INTEGRAL_CAPPED'), port.withConstantValue(0)] })
-            + firstValid.withOutPorts({ output: port.withSignalName('CONCURRENCY_INCREMENT_NORMAL') }),
+          component.withIntegrator(
+            integrator.new()
+            + integrator.withInPorts({
+              input: port.withSignalName('CONCURRENCY_INCREMENT_SINGLE_TICK'),
+              max: port.withSignalName('MAX_EMA'),
```
DariaKunoichi, line 63: not sure what to put here as a max value... should I create constant with max float?

tanveergill: Need to cap it ```2*ACCEPTED_CONCURRENCY```

The previous circuit with looped integral was doing the same

DariaKunoichi: So the max for the whole gradient and optimizer is the same - `NORMAL_CONCURRENCY_LIMIT`, got it.

tanveergill (approved):

PR #1112, opened 2023-01-12
Horizontal Pod Scaler

tanveergill:
### Description of change
* Kubernetes Control Point discovery
  * Agent Group Leader is responsible for running a continuous Kubernetes Control Point discovery against Kubernetes API Server to populate a Kubernetes Control Point Cache
  * The idea is to discover Resources which can be scaled
  * At startup, it finds all Subresource of kind `Scale`. The parents of these resources are scalable resources
  * A watch is added for all scalable resources. The scale status is fetched on Add/Update watch changes. A `notifiers.Trackers` is provided for any other subsystem to subscribe to Kubernetes Control Point Changes
  * FluxNinja Plugin sends a snapshot of Kubernetes Control Point Cache every heartbeat
  * Kubernetes Control Point Message
    ```
    message KubernetesControlPoint {
      string api_version = 1;
      string kind = 2;
      string namespace = 3;
      string name = 4;
    }
    ```
* Pod Scaler
  * `horizontal_pod_scaler` spec
    * `kubernetes_object_selector` to uniquely identify a scalable Kubernetes Object (Control Point) at an Agent Group
    * `scale_reporter` which emits `actual_replicas` and `configured_replicas`
    * `scale_actuator` which takes `desired_replicas`
  * Agent
    * The Agent Group Leader is responsible for running the Autoscaler component
    * It subscribes to the Control Point configured in the component from the Kubernetes Control Point Cache
    * The latest Scale Status for the Control Point is synced to etcd under a new status subtree parallel to config and decisions I.E. `/status/pod_autoscaler/<componentID>`
    * It watches `/decisions/pod_autoscaler/<componentID>` for scale changes from Controller. Scale is updated in Kubernetes if there's a change.
  * Controller
    * There is a `desired_replicas` port for setting replica count via a Signal. This is synced by Controller to a decisions path in etcd path as referred in Agent
    * It signals for completing the feedback loop via ports `actual_replicas` and `configured_replicas`
    * This component takes dynamic config to set dry run mode which disabled actuation (auto-scale) but still completed the feedback loop for testing in "dry-run" mode.
* Implements #864 

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1112)


cmd/aperture-agent/agent.go
```diff
@@ -18,12 +18,20 @@ import (
...
 	"github.com/fluxninja/aperture/pkg/distcache"
 	"github.com/fluxninja/aperture/pkg/entitycache"
+	"github.com/fluxninja/aperture/pkg/etcd/election"
 	"github.com/fluxninja/aperture/pkg/k8s"
 	"github.com/fluxninja/aperture/pkg/log"
 	"github.com/fluxninja/aperture/pkg/otelcollector"
 	"github.com/fluxninja/aperture/pkg/peers"
 	"github.com/fluxninja/aperture/pkg/platform"
 	"github.com/fluxninja/aperture/pkg/policies/flowcontrol"
+<<<<<<< HEAD
```
hasit, line 28: Remnants of a merge

manifests/charts/aperture-agent/templates/rbac.yaml
```diff
@@ -165,6 +165,19 @@ rules:
   - patch
   - update
   - watch
+# auto scale permissions
+- apiGroups: ["*"]
+  resources: ["*"]
```
code-shreyas, line 170: Is this intentional ? I believe the agent is expected to function with limited permissions. 
Having * would mean read access to all the resources.
Usually secrets and configmaps are let out

tanveergill: We are getting read permissions to all objects in order to auto-discover any object that is scalable. Can keep this permission under Kubernetes control point discovery helm flag, cc: @hdkshingala

pkg/discovery/kubernetes/control-point-cache.go
```diff
@@ -0,0 +1,250 @@
...
+	k8sClient          k8s.K8sClient
+	// Set of unique controlPoints
+	controlPoints map[ControlPoint]*controlPointState
+	trackers      notifiers.Trackers
+	ctx           context.Context
+	cancel        context.CancelFunc
+}
+
+// controlPointCache implements the ControlPointStore interface.
+var _ ControlPointStore = &controlPointCache{}
```
code-shreyas, line 95: nit : 
var _ ControlPointStore = (*controlPointCache)(nil)
This one avoids struct creation and reduce runtime memory by small %

pkg/discovery/kubernetes/control-point-cache.go
```diff
@@ -0,0 +1,250 @@
...
+
+// controlPointCache implements the ControlPointCache interface.
+var _ ControlPointCache = &controlPointCache{}
+
+// newControlPointCache returns a new ControlPointCache.
+func newControlPointCache(trackers notifiers.Trackers, k8sClient k8s.K8sClient) *controlPointCache {
```
code-shreyas, line 101: The return parameter can be the interface ControlPointCache

pkg/discovery/kubernetes/control-point-discovery.go
```diff
@@ -0,0 +1,178 @@
...
+	discoveryClient   discovery.DiscoveryInterface
+	dynamicClient     dynamic.Interface
+	election          *election.Election
+}
+
+// Start starts the Kubernetes control point discovery.
+func (cpd *controlPointDiscovery) start() {
+	cpd.ctx, cpd.cancel = context.WithCancel(context.Background())
+
+	cpd.waitGroup.Add(1)
```
code-shreyas, line 51: The waitgroup is not being used ? 
 wait() is not being called

pkg/discovery/kubernetes/control-point-discovery.go
```diff
@@ -0,0 +1,178 @@
...
+					cpd.createResourceEventHandlerFuncs(groupVersionResource, kind),
+				)
+
+				// start controller
+				panichandler.Go(func() {
+					controller.Run(cpd.ctx.Done())
+				})
+			}
+
+			<-cpd.ctx.Done()
```
code-shreyas, line 134: to avoid deadlocks, can we have this wait on the channel surrounded by select statement or use waitgroups here ?

pkg/discovery/kubernetes/control-point-discovery.go
```diff
@@ -0,0 +1,178 @@
...
+		DeleteFunc: func(obj interface{}) {
+			controlPoint := controlPointFromObject(obj)
+			cpd.controlPointStore.Delete(controlPoint)
+		},
+	}
+}
+
+func (cpd *controlPointDiscovery) stop() {
+	cpd.cancel()
+	cpd.waitGroup.Wait()
```
code-shreyas, line 177: Question : 
The intention of stop is to stop all the goroutines or wait until the goroutine work is done ?

pkg/discovery/kubernetes/control-point-cache.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/fluxninja/aperture/pkg/panichandler"
+	"google.golang.org/protobuf/proto"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/scale"
+)
+
+// A ControlPoint is identified by Group, Version, Kind, Namespace and Name.
+type ControlPoint struct {
+	Group     string
+	Version   string
```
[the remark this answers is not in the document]
tanveergill: Because it does not include `Name` and `Namespace`

tanveergill:
For reviewers, following missing files just got added:
```
pkg/policies/controlplane/components/actuators/podautoscaler/pod-autoscaler.go
pkg/policies/controlplane/components/actuators/podautoscaler/scale-actuator.go
pkg/policies/controlplane/components/actuators/podautoscaler/scale-reporter.go
pkg/policies/infra/actuators/podautoscaler/pod-autoscaler.go
```

pkg/discovery/kubernetes/control-point-cache.go
tanveergill: Not really, because need to return an object that implements both `ControlPointCache` and `ControlPointStore` at the [caller](https://github.com/fluxninja/aperture/blob/tgill/k8s-control-point-discovery/pkg/discovery/kubernetes/provide.go#L70)

pkg/discovery/kubernetes/control-point-cache.go
tanveergill: Done

pkg/discovery/kubernetes/control-point-discovery.go
tanveergill: There's actually no need to wait here since resource watcher are running in separate goroutines. Removed this. Actually found a real concurrency issue right above this

pkg/discovery/kubernetes/control-point-discovery.go
tanveergill: The intention is for Fx to detect if a goroutine is stuck at shutdown over [here](https://github.com/fluxninja/aperture/blob/tgill/k8s-control-point-discovery/pkg/discovery/kubernetes/provide.go#L93)

pkg/discovery/kubernetes/control-point-discovery.go
tanveergill: It's being waited on in [stop()](https://github.com/fluxninja/aperture/blob/f0ea05d0ced70abc420e17835c1afac3bf32ebe7/pkg/discovery/kubernetes/control-point-discovery.go#L177) to make sure the goroutine ends gracefully at shutdown

harjotgill (approved):

Issue #871, opened 2022-11-01
Variable Component


tanveergill:
## Description
* Rename Constant to Variable

## Spec
```proto
message Variable {
  message DynamicConfig {
    ConstantValue constant_value = 1;
  }

  message Outs {
    OutPort output = 1;
  }

  Outs out_ports = 1;

  string dynamic_config_key = 2;

  DynamicConfig default_config = 3;
}
```

For ConstantValue message definition refer #870

Issue #870, opened 2022-11-01
Allow Invalid as a Signal Constant


tanveergill:
## Spec
```proto
message InPort {
  oneof value {
    // Name of the incoming Signal on the InPort.
    string signal_name = 1;
    // Constant value to be used for this InPort instead of a signal.
    ConstantValue constant_value = 2;
  }
}

message ConstantValue {
  bool valid = 1;
  double value = 2;
}
```

PR #1122, opened 2023-01-13
Provide dynamically configurable constant values in policy

DariaKunoichi:
### Description of change
Renames Constant to Variable, allows it to be dynamically configured.
ConstantValue itself can now be invalid as well.

Closes: GH-870
Closes: GH-871

##### Checklist

- [x] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1122)


krdln (commented):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1122)* status: 0 of 45 files reviewed, 1 unresolved discussion (waiting on @DariaKunoichi)

___
*[`pkg/policies/controlplane/components/constant.go` line 27 at r1](https://reviewable.io/reviews/fluxninja/aperture/1122#-NLhc3VNC0Xmedu2Y2hn:-NLhc3VNC0Xmedu2Y2ho:b-8z4282) ([raw file](https://github.com/fluxninja/aperture/blob/0e2a7de409b77da9e091d7ddf131415f5c385caf/pkg/policies/controlplane/components/constant.go#L27)):*
> ```Go
> 
> // NewVariable creates a variable component.
> func NewVariable(value float64) runtime.Component {
> ```

Nit: This function could still be named `NewConstant`. The fact that it now uses `Variable` under the hood is only implementation detail (note that it returns interface`runtime.Component`).



_Code quote:_
```Go
NewVariable
```

api/aperture/policy/language/v1/policy.proto
```diff
@@ -969,19 +969,31 @@ message PromQL {
   }]; // @gotags: default:"10s"
 }
 
-// Component that emits a constant value as an output signal
-message Constant {
-  // Outputs for the Constant component.
+// Constant value that can be used instead of a signal in ports. Can be set to be invalid.
+message ConstantValue {
+  bool valid = 1;
```
harjotgill, line 974: This is making it a bit ugly to use. Maybe InvalidValue can be a separate message?

DariaKunoichi: Ok, let's ask @tanveergill 
Right now I created it as stated in [task description](https://github.com/fluxninja/aperture/issues/870)

tanveergill: @DariaKunoichi,
Perhaps we can reverse the boolean to `invalid` so that providing only a double `value` still works. Can make it a `oneof` `invalid` or `value`

tanveergill (approved):
Okay to merge once spec is simplified:
```
message ConstantValue {
  oneof const {
    bool invalid = 1;
    double value = 2;
  }
}
```

krdln (commented):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1122)* status: 0 of 64 files reviewed, 2 unresolved discussions (waiting on @DariaKunoichi and @harjotgill)

___
*[`api/aperture/policy/language/v1/policy.proto` line 974 at r1](https://reviewable.io/reviews/fluxninja/aperture/1122#-NLfkM48KOCbJcpPAtW9-r1-974:-NM87jya31FlkjlAGrUx:b-5nqwb) ([raw file](https://github.com/fluxninja/aperture/blob/0e2a7de409b77da9e091d7ddf131415f5c385caf/api/aperture/policy/language/v1/policy.proto#L974)):*
<details><summary><i>Previously, tanveergill (Tanveer Gill) wrote…</i></summary><blockquote>

@DariaKunoichi,
Perhaps we can reverse the boolean to `invalid` so that providing only a double `value` still works. I.E.:
```
message ConstantValue {
  oneof const {
    bool invalid = 1;
    double value = 2;
  }
}
```
</blockquote></details>

:+1: for `bool invalid` so that default is easier to work with.

BTW, do we need explicit `bool invalid`? Right now `constant_value: "NaN"` should work already.
___
*[`api/aperture/policy/language/v1/policy.proto` line 982 at r2](https://reviewable.io/reviews/fluxninja/aperture/1122#-NM88zll0LpGnzntptMA:-NM88zll0LpGnzntptMB:bm55li2) ([raw file](https://github.com/fluxninja/aperture/blob/8141ae3b8f42206bdd32c3545cfea3cd4eb4d466/api/aperture/policy/language/v1/policy.proto#L982)):*
> ```protobuf
> 
> // Component that emits a constant value as an output signal
> message Constant {
> ```

Btw, we could've kept the `Constant` message and "component" and translate it internally to `Variable`. This way we could avoid any changes in configs.

api/aperture/policy/language/v1/policy.proto
tanveergill: @krdln, @DariaKunoichi,
I am not sure how that would work. Proto Unmarshal would expect a field to be a concrete type. How can the same field be decoded as float or string. Maybe I am missing something?

DariaKunoichi: I did not know that it will work. But here is a test from go-playground:
https://go.dev/play/p/6lQDfLOHz4_c

And this is excerpt from docs:
> ParseFloat recognizes the string "NaN", and the (possibly signed) strings "Inf" and "Infinity" as their respective special floating point values. It ignores case when matching.

tanveergill: Using a string to represent a floating point in the underlying proto message is very misleading. Discussed this with Harjot, and we feel that a dedicated field is needed to represent special numbers instead of a bool in the current proposal. The problem with representing invalid as a oneof is that setting invalid to false is an ambiguous state since value if undefined. Here is the final proposal:

```
message ConstantValue {
  oneof const {
    string special_value = 1; // @gotags: validate:"oneof=NaN +Inf -Inf"
    double value = 2;
  }
}
```

tanveergill: Let's rename `ConstantValue` to `ConstantSignal` to keep a consistent terminology

DariaKunoichi: ok, will do

PR #1136, opened 2023-01-16
Fix Control Point Cache provide

kwapik:
### Description of change
* Move `ProvideControlPointCache` from `Invoke` to `Provide`.
* Make `ControlPointCache` dependency optional in FN plugin, as Controller won't use it.

##### Checklist

- [ ] Breaking changes


krdln (approved):
Approving as it's a fix, and k8s autoscaler is not yet used afaik, but I feel something is inverted wrt Invoke/Provide.

PR #1141, opened 2023-01-16
Add default channel for Load Shed Event alert

kwapik:
### Description of change
In order to properly process an Alert in our system we need to have at least one Alerts Channel assigned to it. This change introduces default Alerts Channels for Load Shed Event Alerts.

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1141)


kwapik:
Decided to drop the requirement to have at least one alert channel.

Issue #1150, opened 2023-01-17
Streamline default metrics collection in Kubernetes


tanveergill:
## What
* Configure [Kubelet Stats Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kubeletstatsreceiver) to gather K8s infra metrics
* See how to expose TLS config if required

Issue #1123, opened 2023-01-13
Passthrough and merge user defined OTEL config into Aperture Agent


tanveergill:
## Why
* Collect metrics using ~~Importers and Processors~~ Receivers available in OTEL collector
* Write PromQL queries in Aperture Policies to generate Signals from these metrics
* Use these Signals for infra automation

## What/How
* A [slice of OTEL configs](https://github.com/fluxninja/aperture/blob/b79bea4dc0739d0a49dcd89bf03a32070d3e04ee/pkg/otelcollector/otelcollector.go#L46) is gathered by [otelcollector.setup](https://github.com/fluxninja/aperture/blob/b79bea4dc0739d0a49dcd89bf03a32070d3e04ee/pkg/otelcollector/otelcollector.go#L50) 
* Refer how an instance of OTEL config is provided by [fluxninja plugin](https://github.com/fluxninja/aperture/blob/main/plugins/service/aperture-plugin-fluxninja/otel/provide.go#L50)
* Use the same technique as above to pass used defined OTEL config
* Append exporters at the end of user defined pipelines in order to export those metrics to Aperture's Prometheus
* Add user defined pipelines to [existing OTEL config in Aperture](https://github.com/fluxninja/aperture/blob/b79bea4dc0739d0a49dcd89bf03a32070d3e04ee/pkg/otelcollector/config.go#L240), corresponding [docs reference](https://docs.fluxninja.com/docs/references/configuration/agent#otel-config)

PR #1173, opened 2023-01-20
Collect Kubernetes metrics

kwapik:
### Description of change

Ref: GH-1150

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1173)


plugins/service/aperture-plugin-fluxninja/otel/provide.go
```diff
@@ -128,9 +129,13 @@ func addFNToPipeline(
 
 func addMetricsSlowPipeline(baseConfig, config *otelcollector.OTELConfig) {
 	addFluxninjaPrometheusReceiver(baseConfig, config)
+	addFluxninjaKubeletStatsReceiver(baseConfig, config)
```
tanveergill, line 132: Why is this in the fluxninja plugin? It's better to pass this config explicitly via [user visible config](https://github.com/fluxninja/aperture/issues/1123) so that the user can enable the precise metrics they need. We would enable only a select few metrics by default

kwapik: The part in fluxninja plugin is not defining this pipeline from scratch, but it is duplicating the original pipeline with two changes:
1. Increased scrape interval from `1s` to `10s` as we don't need such granularity in cloud metrics
2. Added fluxninja exporter which is writing metrics to the cloud

I can wait for the user defined configs to be completed and modify this PR to use this.

kwapik:
Closing in favor of #1227

PR #1220, opened 2023-01-31
Upgrade olric

hasit:
### Description of change
- Refactor stats grpc service to work with structpb

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1220)


pkg/distcache/stats-service.go
```diff
@@ -37,21 +42,29 @@ func (svc *DistCacheService) GetStats(ctx context.Context, _ *emptypb.Empty) (*d
...
+	return structpbStats, nil
 
-	// Removing empty partitions to reduce the response size
-	partitions := make(map[uint64]*distcachev1.Partition)
-	for key, partition := range newStats.Partitions {
-		if partition.Length != 0 {
-			partitions[key] = partition
-		}
-	}
+	// newStats := &distcachev1.Stats{}
```
hdkshingala, line 53: @hasit This logic was added as the actual response from olric stats endpoint contains a lot of partitions with length = 0 (> 250), which makes the response size too big.

As we are using the structpb directly, we can use the same logic on the original stats object to remove the empty partitions so that the response size remains small and readable. Please share your views on this.

hasit: Yes, I am aware and going to bring it back.

hasit: [32029b0](https://github.com/fluxninja/aperture/pull/1220/commits/32029b0bfbf23a16ada5e5137e6f20928246b23e) addresses this.

hdkshingala (approved):

harjotgill (approved):

Issue #1147, opened 2023-01-17
Alerter in Status Registry


tanveergill:
## Why
* Status Registry is a hierarchical system. Any component can derive Child registries by specifying a Key for the Child. The derived Registry maintains the context of its heredity which applies to:
  * The Status URI: <parent_uri>/<child_sub_path>
  * Logger: Appends key:value of the child registry to the existing key:value's from the parent
* Similar to Status and Logger, it would be useful to make an Alerter available in the Registry. Using the Alerter, developers can create Alerts for any sub-system with all the context about its heredity intact.

## What
* Introduce `GetAlerter` method in `Registry` interface
* The Alerter at each Registry needs to be labelled with key:value of the Registry and all its parents
* Change the existing methods for creating Child registries to take Key and Value strings instead of just a Key
  * Example Status before:
    * URI: `/concurrency_limiter/agent_group-default-policy-service1-demo-app-component_index-13/fx_runner`
    * Labels: `root:concurrency_limiter`, `concurrency_limiter:agent_group-default-policy-service1-demo-app-component_index-13`, `agent_group-default-policy-service1-demo-app-component_index-13:fx_runner`
  * Example Status after:
    * URI: `/component/concurrency_limiter/component_id/agent_group-default-policy-service1-demo-app-component_index-13/sub_system/fx_runner`
    * Labels: `component:concurrency_limiter`, `component_id:agent_group-default-policy-service1-demo-app-component_index-13`, `sub_system:fx_runner`
* Generate an Alert whenever `SetStatus` has a non-nil error. Set error string as the Alert name, the Status URI as a label in addition to all other labels in the Status context

PR #1226, opened 2023-01-31
Add alerter to status registry

DariaKunoichi:
### Description of change
Ref: GH-1147
##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1226)


pkg/status/registry.go
```diff
@@ -21,10 +25,17 @@ type Registry interface {
 	Key() string
 	HasError() bool
 	GetLogger() *log.Logger
+	GetAlerter() alerts.Alerter
 }
 
 var _ Registry = &registry{}
 
+const (
+	alertChannel = "status_registry"
```
DariaKunoichi, line 34: @tanveergill I hardcoded this because status registry is for our internal use
We do not configure alert channels in alerter itself (it's not the same as alerter component) - https://github.com/fluxninja/aperture/blob/main/pkg/alerts/provide.go#L15
I can change `GetAlerter` method to also specify the channel list if you want

krdln (approved):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1226)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @DariaKunoichi)

___
*[`pkg/status/registry.go` line 82 at r2](https://reviewable.io/reviews/fluxninja/aperture/1226#-NNC8fAn96F9u3J64-9Y:-NNC8fAn96F9u3J64-9Z:b-hbheol) ([raw file](https://github.com/fluxninja/aperture/blob/c77ab979ecc713aec7f19e027335f8aa756851cc/pkg/status/registry.go#L82)):*
> ```Go
> 			children: make(map[string]*registry),
> 			logger:   r.logger.WithStr(r.key, key),
> 			alerter:  r.alerter,
> ```

I understand that this will get some labels in subsequent PR? Because right now all alerters will have `uri: /`

DariaKunoichi (commented):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1226)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @DariaKunoichi and @krdln)

___
*[`pkg/status/registry.go` line 82 at r2](https://reviewable.io/reviews/fluxninja/aperture/1226#-NNC8fAn96F9u3J64-9Y:-NNCAZjjE9ee6eoj5_PD:b-2ecmn5) ([raw file](https://github.com/fluxninja/aperture/blob/c77ab979ecc713aec7f19e027335f8aa756851cc/pkg/status/registry.go#L82)):*
<details><summary><i>Previously, krdln (Michał Krasnoborski) wrote…</i></summary><blockquote>

I understand that this will get some labels in subsequent PR? Because right now all alerters will have `uri: /`
</blockquote></details>

yes, exactly. I don't have KV here yet so stub labels are used for now

PR #1285, opened 2023-02-04
Refactor aperturectl

hasit:
### Description of change
This command allows users to apply raw policies from file/dir created by `generate` command.

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1285)


harjotgill (approved):

hasit:
Todo:
1. Make `generate --apply` use the same functions as `apply` command. Reduce code repetition. This would entail moving `kube-config` to a global level
2. Update `apply` command
- Add `apply policies`
- Add `apply dynamic-config`
3. `latest` tag is not created as `latest`, instead as the reference fetched from remote. Put the latest pulled version under `latest` so that passing `--version=latest` to other commands works
4. Pull should overwrite everything under the version being pulled

PR #1343, opened 2023-02-10
Update OTEL to v0.71.0

kwapik:
### Description of change
This updates OpenTelemetry Collector, FN fork of it and contrib packages to latest released version, which is `v0.71.0`

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1343)


jmichalak-fluxninja (approved):

krdln (approved):
Note to self: benchmark after merging and see if we haven't introduced O(n²) slowdown in `key` (and whether that bothers us).

kwapik:
image

krdln:
@kwapik Looked at the profile. `key` takes about 2% – just as before. The `Get()` in the loop, that's possible quadratic takes `0.27%`, so it's not a problem now, but it's "on the edge". Eg. if we have 10× more labels, we'd have 100× slowdown and `Get()` would be at ~27%. Worth addressing in the future.

PR #1350, opened 2023-02-13
Propagate `instance` in user defined metrics

kwapik:
### Description of change
This label is required to avoid metrics with same set of labels and timestamp to be pushed to FluxNinja ARC. It causes errors in the metrics backend.

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1350)


hdkshingala (approved):

kwapik:
image

PR #1356, opened 2023-02-14
Use Tristate bool readings in all components that deal with and/or logic 

karansohi:
In all components that were modified, I switched the component.Valid() boolean comparison to the tristate boolean comparison

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1356)


pkg/policies/controlplane/runtime/tristate/tristate.go
```diff
@@ -49,11 +49,11 @@ func (b Bool) ToReading() runtime.Reading {
 	}
 }
 
-// FromReading interprets runtime.Reading as Bool, mapping 0 to False, any
+// ReadValue interprets runtime.Reading as Bool, mapping 0 to False, any
 // valid non-zero to True and Invalid to Unknown.
 //
 // (It's the same mapping as in ToReading, but allowing more truthy values).
-func FromReading(reading runtime.Reading) Bool {
+func ReadValue(reading runtime.Reading) Bool {
```
harjotgill, line 56: Why rename this function?

pkg/policies/controlplane/components/switcher.go
```diff
@@ -39,7 +40,7 @@ func (dec *Switcher) Execute(inPortReadings runtime.PortToReading, tickInfo runt
 
 	var output runtime.Reading
 
-	if switchValue.Valid() && switchValue.Value() != 0 {
+	if tristate.ReadValue(switchValue) > 0 {
```
harjotgill, line 43: This looks very incorrect. Earlier we were reading actual value. Tristate "ReadValue" (renamed recently) is not reading value but instead, it's telling whether the reading is valid or not. Is this a correct behavior?
We also need to compare against tristate.Bool -- i.e. understand Unknown/true/false values. This comparison will cause the program to enter the if condition when tristate is Unknown and that was not the case earlier -- earlier we returned from this Execute method when the reading was invalid.

harjotgill (changes requested):

pkg/policies/controlplane/runtime/tristate/tristate.go
karansohi: Restored to previous name

pkg/policies/controlplane/components/switcher.go
karansohi: shifted to tristate.True

krdln (changes requested):
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1356)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @harjot, @harjotgill, and @karansohi)

___
*[`pkg/policies/controlplane/circuitfactory/component-factory_test.go` line 183 at r2](https://reviewable.io/reviews/fluxninja/aperture/1356#-NOJyLFe9s-ZllvLQQtN:-NOJyLFe9s-ZllvLQQtO:by72hfg) ([raw file](https://github.com/fluxninja/aperture/blob/6c95cc8aa03dc35e0f7c6a798790e3cf63ccb1fa/pkg/policies/controlplane/circuitfactory/component-factory_test.go#L183)):*
> ```Go
>       Expect(options).To(BeNil())
>     })
>   })
> ```

Nit/sidenote: I kinda think these component-factory tests are kinda redundant. A test for running components should cover their creation too. See eg. `holder_test.go`.
___
*[`pkg/policies/controlplane/components/alerter.go` line 74 at r2](https://reviewable.io/reviews/fluxninja/aperture/1356#-NOJwwBaC311qa3D8NMH:-NOJwwBaC311qa3D8NMI:buribtl) ([raw file](https://github.com/fluxninja/aperture/blob/6c95cc8aa03dc35e0f7c6a798790e3cf63ccb1fa/pkg/policies/controlplane/components/alerter.go#L74)):*
> ```Go
> 	signalValue := inPortReadings.ReadSingleReadingPort("signal")
> 
> 	if tristate.FromReading(signalValue) > tristate.True {
> ```

This condition is never true. Did you mean `== tristate.True`? Feel free to add helper methods like `IsTrue` on `tristate.Bool`.

BTW. Comparing with `>` is very weird... Perhaps we should make `tristate.Bool` a struct with private field, so that such operator won't compile.
___
*[`pkg/policies/controlplane/components/integrator.go` line 52 at r2](https://reviewable.io/reviews/fluxninja/aperture/1356#-NOJxWHd6_8LVd3f9GaM:-NOJxWHd6_8LVd3f9GaN:bs4q9op) ([raw file](https://github.com/fluxninja/aperture/blob/6c95cc8aa03dc35e0f7c6a798790e3cf63ccb1fa/pkg/policies/controlplane/components/integrator.go#L52)):*
> ```Go
> 	inputVal := inPortReadings.ReadSingleReadingPort("input")
> 	resetVal := inPortReadings.ReadSingleReadingPort("reset")
> 	if tristate.FromReading(resetVal) > tristate.True {
> ```

Ditto.
___
*[`pkg/policies/controlplane/components/switcher.go` line 43 at r1](https://reviewable.io/reviews/fluxninja/aperture/1356#-NOCEsB3do4cDdPtdorn-r1-43:-NOJxmHz0S1voJmKSmad:b-3kb0sg) ([raw file](https://github.com/fluxninja/aperture/blob/6e466726b5765fbf156476e7f7f0460f63b64bfe/pkg/policies/controlplane/components/switcher.go#L43)):*
<details><summary><i>Previously, karansohi (Karanbir Singh) wrote…</i></summary><blockquote>

shifted to tristate.True
</blockquote></details>

@harjot, should we make switcher output Invalid when switchValue is invalid?

(In such case, I think here we can use switch case (`switch tristate.FromReading(switchValue)` with
cases for unknown/true/false. This way we'll explicitly cover all variants.

pkg/policies/controlplane/runtime/tristate/tristate.go
```diff
@@ -63,6 +63,21 @@ func FromReading(reading runtime.Reading) Bool {
 	return True
 }
 
+// IsTrue returns the tri-state boolean True.
+func IsTrue() Bool {
```
harjotgill, line 67: These methods don't make much sense. What Michal was saying that add helper methods for Bool Type. e.g. func (bool Bool) IsTrue()

karansohi: Done. I also added isFalse and isUnkown because if it is not true we have to check if it is either false or unknown.

pkg/policies/controlplane/runtime/circuit_test.go
```diff
@@ -83,3 +83,62 @@ var _ = Describe("Circuit", func() {
...
 	})
 })
+
+var _ = Describe("Integrator", func() {
+	It("integrates signals", func() {
+		circuit, err := sim.NewCircuitFromYaml(
+			`
+      evaluation_interval: 1s
+      components:
+      - integrator:
```
harjotgill, line 93: Merge latest main. I changed integrator a bit.

harjotgill (changes requested):

pkg/policies/controlplane/runtime/circuit_test.go
harjotgill: Also, integrator had lot of bugs -- e.g. min/max not doing what was supposed to, absence of min or max would give wrong result and so on. I fixed some bugs yesterday and fixing more today. 
As a learning, we need our test cases to cover these corner cases as this PR could have caught many issues as well. cc: @DariaKunoichi

DariaKunoichi: @harjotgill thanks for catching it! I'll work on adding more tests in my future tasks

krdln (approved):
> Reviewed 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1356)* status: 1 of 7 files reviewed, 4 unresolved discussions (waiting on @DariaKunoichi, @harjot, and @harjotgill)

harjotgill (approved):

DariaKunoichi (approved):

PR #1365, opened 2023-02-15
increase `PV` to `10Gi` for aperture controller

slayer321:
### Description of change
We are getting `Kubernetes Volume full in four days` alerts on #producation-alert slack channel . After looking into this we found that the average usage is around 6 GB and the current Capicity was set to 8 GB . After having discussion with Szymon . We decided to Increase to Volume to 10 Gi.

cc @sbienkow-ninja 

Screenshot from 2023-02-14 22-46-15

##### Checklist

- [ ] Breaking changes

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1365)


slayer321:
This issue was resolved in this PR https://github.com/fluxninja/argo-manifests/pull/76

PR #1474, opened 2023-02-28
Generate dashboards from blueprints in playground

hdkshingala:
### Description of change

Grafana dashboards will be generated using the same policy blueprints bundle instead of static one from `lib/apps`.

##### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/fluxninja/aperture/1474)

### DIFF
--- a/playground/scenarios/demo-app/metadata.json
+++ b/playground/scenarios/demo-app/metadata.json
@@ -6,7 +6,9 @@
     {
       "blueprint_name": "policies/latency-aimd-concurrency-limiting",
       "policy_name": "service1-demo-app",
-      "values_file": "policies/service1-demo-app.yaml"
+      "values_file": "policies/service1-demo-app.yaml",
+      "dashboard_name": "service1-demo-app-dashboard",
+      "dashboard_mixin_dir": "dashboards/"
     }
   ],
   "images": [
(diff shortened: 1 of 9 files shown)


harjotgill (approved):

playground/scenarios/demo-app/metadata.json
```diff
@@ -6,7 +6,9 @@
     {
       "blueprint_name": "policies/latency-aimd-concurrency-limiting",
       "policy_name": "service1-demo-app",
-      "values_file": "policies/service1-demo-app.yaml"
+      "values_file": "policies/service1-demo-app.yaml",
```
tanveergill, line 9: What about signals dashboard? How is that generated?

hdkshingala: Signals dashboard is still coming from `lib/apps` as we don't have it in scenarios. So basically it will always be there when we run playground.

tanveergill: Okay, the panels in it are filtered by policy name. We should either make policy one of the variables so that it works for all policies or generate it as part of scenario

hdkshingala: Okay, I'll check it tomorrow and most probably will generate as part of the scenarios.

hdkshingala: Handled this as part of https://github.com/fluxninja/aperture/pull/1500

PR #1536, opened 2023-03-07
Set FN exporter `num_consumers` to 1

kwapik:
### Description of change
Having more than one queue consumer lead to metrics batches being sent out of order and thus being rejected by FN ARC.

##### Checklist


harjotgill (approved):

kwapik:
Follow up in upstream https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/19442

PR #1567, opened 2023-03-10
Fix brew release flow

sbienkow-ninja:
### Bug Fix
- Fixes the brew release flow by changing the executor, adding a commit message, and creating a pull request using the GitHub CLI.

### DIFF
--- a/.circleci/post-release.yaml
+++ b/.circleci/post-release.yaml
@@ -209,7 +209,7 @@ jobs:
         type: string
         description:
           Deployment key to use when interacting with remote repository
-    executor: ubuntu-18-executor
+    executor: python-cimg-executor
     environment:
       TAP_REPO: /home/circleci/taps
       APERTURE_REPO: /home/circleci/project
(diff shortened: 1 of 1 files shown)


.circleci/post-release.yaml
```diff
@@ -243,20 +243,21 @@ jobs:
...
             ./scripts/update_brews.py update
 
             git status
             git diff
             git add .
-            git commit -m "Update brews for release ${VERSION}"
+            git commit -m "${msg}"
             git push origin "${branch}"
-            # gh pr create --title "${MSG}" --body "" --label "pr-pull"
+            gh pr create --title "${msg}" --body "" --label "pr-pull"
```
hdkshingala, line 260: I think I tried this and it is not triggering the release on brew so should we merge directly to main so that we don't have to manually merge this PR?

sbienkow-ninja: We were missing the label earlier. And this might fix it too: https://github.com/fluxninja/homebrew-aperture/pull/20

hdkshingala (approved):

PR #1573, opened 2023-03-13
Replace dynamic plugins with statically linked extensions

harjotgill:
### Description of change

- Plugins removed and replaced by extensions
- Added extension builder in aperturectl

##### Checklist

- [x] Breaking changes


kwapik (approved):

cmd/aperture-agent/Dockerfile
```diff
@@ -1,40 +1,35 @@
...
+ENV APERTURECTL_BUILD_VERSION=${APERTURECTL_BUILD_VERSION}
+ARG APERTURECTL_BUILD_GIT_COMMIT_HASH
+ENV APERTURECTL_BUILD_GIT_COMMIT_HASH=${APERTURECTL_BUILD_GIT_COMMIT_HASH}
+ARG APERTURECTL_BUILD_GIT_BRANCH
+ENV APERTURECTL_BUILD_GIT_BRANCH=${APERTURECTL_BUILD_GIT_BRANCH}
+ARG APERTURECTL_BUILD_BUILDER_CONFIG_FILE
+ENV APERTURECTL_BUILD_BUILDER_CONFIG_FILE=${APERTURECTL_BUILD_BUILDER_CONFIG_FILE}
+
+# If config file exists, copy it to the build context
+RUN if [[ -f "${APERTURECTL_BUILD_BUILDER_CONFIG_FILE}" ]]; then \
```
kklimonda-fn, line 18: This is failing in CI as the script is executed with sh, and not bash:

```
#15 [builder 4/5] RUN if [[ -f "" ]]; then   cp "" ./config.yaml;   else   echo "APERTURECTL_BUILD_BUILDER_CONFIG_FILE is not set";   touch ./config.yaml;   fi
#15 sha256:52c4d7e69d01cb46e1b52bfe5aac7d81133d69ebf64378ef70c6e131615537f8
#15 0.322 /bin/sh: 1: [[: not found
#15 0.323 APERTURECTL_BUILD_BUILDER_CONFIG_FILE is not set
#15 DONE 0.3s
```

kklimonda-fn (changes requested):

hasit:
I still believe there is no way to pass files during runtime except for attaching a new volume. We should prepare the file, include it in the docker context, and then use it. We can still continue to accept config file name.

kklimonda-fn (approved):

PR #1621, opened 2023-03-24
Refactor installation mode

kwapik:
### Description of change
This removes whole logic of determining installation mode.

Rationale:
1. Previous solution was not working
2. If we are going to rely on some files/envs being set, we can just have installation mode as a config option.

TODO:

- [ ] Actually set this config option on installation

##### Checklist

- [ ] Breaking changes


extensions/fluxninja/heartbeats/provide.go
harjotgill, line 1: @openai several functions to detect the runtime platform (e.g. kubernetes daemonset etc) were removed as these functions didn't work. Do you have any suggestions on how to detect the underlying platform the process is running, without passing it as an environment variable?

harjotgill: I am interested in detecting whether the process runs as a Kubernetes Dameonset or a Sidecar container. If it's not running in Kubernetes at all,  I would like to detect whether it's running on Linux or some other OS (unknown). Please generate the code if possible.

kwapik: @openai I couldn't find a file with pods details like the one you're reading in the snippet above. I'm using kubernetes version `v1.23.16`.

kwapik: So there is no file like you mentioned above and I need to use K8s API instead?

kwapik: I don't see any `daemonset` label on my DaemonSets in K8s. Who is responsible for adding them?

kwapik: So, if I need to add labels like `daemonset` or env variables like `POD_NAME` and `POD_NAMESPACE` anyway, I can just have one `INSTALLATION_MODE` env, without the additional K8s API calls.

kwapik:
image

operator/controllers/utils.go
```diff
@@ -215,6 +215,10 @@ func AgentEnv(instance *agentv1alpha1.Agent, agentGroup string) []corev1.EnvVar
 			Name:  "APERTURE_AGENT_SERVICE_DISCOVERY_KUBERNETES_ENABLED",
 			Value: "false",
 		})
+		envs = append(envs, corev1.EnvVar{
```
hdkshingala, line 218: We should update this value in the actual CR instance as well so that we don't end up having 2 different values, one set by defaulter webhook and one passed using environment variable.

hdkshingala: We update some CR instance fields in reconciliation loop already and make update call at the end of the loop

harjotgill (approved):

PR #1634, opened 2023-03-28
Add control point type

sbienkow-ninja:
### Description of change

Add Type field to FlowControlPoint.

##### Checklist

- [ ] Breaking changes

### Summary by OpenAI

Release Notes:

- New Feature: A "Type" field has been added to the `FlowControlPoint` message, improving type safety and maintainability.
- Refactor: Multiple files have been updated to use a new typed control point ID instead of the old one.

> "Typed control points galore, 
> Type safety we can't ignore. 
> With this PR, our code's more sound, 
> Bugs and errors won't stick around."


cmd/aperture-agent/extensions.go
```diff
@@ -1,19 +1,21 @@
 package main
 
 import (
+	"github.com/fluxninja/aperture/extensions/fluxninja"
```
harjotgill, line 4: This should be revered

harjotgill (dismissed):

.circleci/continue-workflows.yml
```diff
@@ -216,7 +216,7 @@ jobs:
         type: string
         description: working directory
     executor: go-cimg-executor
-    resource_class: large
+    resource_class: xlarge
```
krdln, line 219: Intended?

Makefile
```diff
@@ -168,6 +168,7 @@ operator-generate: ## Generate code containing DeepCopy, DeepCopyInto, and DeepC
 	$(CONTROLLER_GEN) object:headerFile="operator/hack/boilerplate.go.txt" paths="./cmd/..."
 	$(CONTROLLER_GEN) object:headerFile="operator/hack/boilerplate.go.txt" paths="./pkg/..."
 	$(CONTROLLER_GEN) object:headerFile="operator/hack/boilerplate.go.txt" paths="./operator/..."
+	$(CONTROLLER_GEN) object:headerFile="operator/hack/boilerplate.go.txt" paths="./extensions/..."
```
krdln, line 171: Intended?

pkg/policies/flowcontrol/selectors/control-point-id.go
```diff
@@ -10,19 +10,36 @@ import (
...
 // protobuf-generated struct cannot be used as map keys.
 type GlobalControlPointID struct {
+	TypedControlPointID
+	AgentGroup string
+}
+
+// UntypedGlobalControlPointID is just like GlobalControlPointID but embedding the ControlPointID instead of TypedControlPointID.
+//
+// Useful for defining a control point to find, without having to specify the source.
+type UntypedGlobalControlPointID struct {
```
krdln, line 42: Nit: This assymetry in naming is a bit weird (global is "by default" Typed and local is by default "Untyped").

krdln (approved):
LGTM, but please verify if you wanted to commit changes to makefile and circle config

.circleci/continue-workflows.yml
sbienkow-ninja: Yes - without this I noticed the coverage job die after using all 8GB of memory (actual usage was 7.98GB). Now we are using around 10GB of memory. We could probably either bump up the proc count or decrease the proc count and resource class instead.

Makefile
sbienkow-ninja: Yes, the extensions still had old plugin generation before (mostly old comments). I believe we do want to keep this.

pkg/policies/flowcontrol/selectors/control-point-id.go
sbienkow-ninja: I know. That untyped global was added in the end. But I think it's worth refactoring to be consistent.

sbienkow-ninja: Updated.

PR #1687, opened 2023-04-04
Fix panic in leader-only receiver Shutdown

krdln:
##### Checklist


krdln:
(btw, this is pretty harmless, as it's panic on shutdown and this receiver is not enabled by default and also it doesn't occur always)

PR #1798, opened 2023-04-24
Added replicaCount flag for namespace scoped installation

hdkshingala:


kklimonda-fn (commented):
Do we need some sort of leadership logic on the agent side to make sure that only one agent is handling all requests, and other agents are forwarding requests to the leader?

hdkshingala:
> Do we need some sort of leadership logic on the agent side to make sure that only one agent is handling all requests, and other agents are forwarding requests to the leader?

As we have service discovery disabled in this mode, dist_cache shouldn't be required. Each agent should be able to execute requests individually but let me confirm it in the dev channel.

kklimonda-fn (approved):

PR #1909, opened 2023-05-09
Blueprint for Auto-scale

### LINKED ISSUE #1647: Blueprints for auto-scaling
## What
* Blueprints hierarchy
  * auto-scale/base
    * What:
      * Not a real blueprint but rather shared code for the variants.
      * There is a function that takes a ScalingBackend and generates the policy.
      * The base policy expects the variants to define the ScalingBackend.
    * Policy:
      * Constructs `AutoScaler` component without the Scaler.
      * Config:
        * promql_scale_out_controllers (new schema): List of scale out controllers.
          * Attributes: Query string, threshold, GradientParameters with sensible defaults
          * Constructed as ScaleOutController.IncreasingGradient with signal from PromQL and constant threshold
        * promql_scale_in_controllers: Similar as above but for scale in.
        * scaling_parameters (refers to ScalingParameters schema in policy proto)
  * auto-scale/pod-auto-scaler
    * What:
      * Builds upon base. Invokes the policy function with KubernetesReplicas scaling backend
      * Config:
        * Inherits base config plus below
        * kubernetes_replicas (refers to KubernetesReplicas schema in policy protocol)

hdkshingala:
Fixes #1647 

### Summary by OpenAI

**New Feature:**
- Added auto-scaling module with default values, Grafana dashboard, and PodAutoScaler blueprint

**Documentation:**
- Created new documentation files for auto-scaling blueprints and updated sidebar positions

**Refactor:**
- Updated `policies.libsonnet` to import `PodAutoScaler`

> 🎉 Auto-scaling takes the stage, 🚀
> A new feature we proudly engage. 🌟
> With docs and code refined, ✍️
> Our users' needs in mind. 💡
> Let's celebrate this milestone, 🥳
> As our software further has grown! 📈

### DIFF
--- a/blueprints/policies/auto-scaling/base/config-defaults.libsonnet
+++ b/blueprints/policies/auto-scaling/base/config-defaults.libsonnet
@@ -1,0 +1,78 @@
+local auto_scaling_base_defaults = {
+  policy_name: '__REQUIRED_FIELD__',
+
+  components: [],
+
+  resources: {
+    flow_control: {
+      classifiers: [],
+    },
+  },
+
+  evaluation_interval: '1s',
+};
+
+local promql_scale_out_controller_defaults = {
+  query_string: '__REQUIRED_FIELD__',
+  threshold: 1.0,
+  gradient: {
+    slope: -1.0,
+  },
+};
+
+local promql_scale_in_controller_defaults = {
+  query_string: '__REQUIRED_FIELD__',
+  threshold: 0.5,
+  gradient: {
+    slope: 1.0,
+  },
+};
+
+local scaling_parameters_defaults = {
+  scale_in_cooldown: '40s',
+  scale_out_cooldown: '30s',
+};
+
+local auto_scaling_defaults = auto_scaling_base_defaults {
+  promql_scale_out_controllers: [
+    promql_scale_out_controller_defaults,
+  ],
+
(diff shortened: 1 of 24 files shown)


blueprints/policies/auto-scaling/base/config-defaults.libsonnet
```diff
@@ -0,0 +1,78 @@
...
+  gradient: {
+    slope: -1.0,
+  },
+};
+
+local promql_scale_in_controller_defaults = {
+  query_string: '__REQUIRED_FIELD__',
+  threshold: 0.5,
+  gradient: {
+    slope: 1.0,
```
tanveergill, line 27: Slope defines the direction of correction i.e. whether the scale should increase/decrease when signal is greater than set point. Default slope should be the same for both scale in and scale out controllers. The default slope is 1 for both in the proto spec. This works well for utilization metrics, e.g. if cpu is greater than threshold then scale out

tanveergill (approved):

blueprints/policies/auto-scaling/base/config-defaults.libsonnet
hdkshingala: Got it. Updated the value to 1.0. Thanks!

PR #2146, opened 2023-06-13
Do `go mod download` in separate layer

kwapik:
### Description of change
This should speed up building of Agent and Controller.

##### Checklist

### Summary by OpenAI

**Chore:**
- Speed up Agent and Controller Docker build process by adding a separate layer for `go mod download`

> 🚀 Faster builds, we celebrate,
> 🎉 With layers separate,
> 🏗️ Agent and Controller now create,
> ⏱️ A Docker image at a quicker rate!

### DIFF
--- a/cmd/aperture-agent/Dockerfile
+++ b/cmd/aperture-agent/Dockerfile
@@ -2,6 +2,10 @@
 FROM golang:1.20.4-bullseye AS builder
 
 WORKDIR /src
+
+COPY go.mod go.sum ./
+RUN go mod download
+
 COPY --link . .
 
 # App build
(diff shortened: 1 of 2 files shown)


cmd/aperture-agent/Dockerfile
```diff
@@ -2,6 +2,10 @@
 FROM golang:1.20.4-bullseye AS builder
 
 WORKDIR /src
+
+COPY go.mod go.sum ./
+RUN go mod download
```
krdln, line 7: This will slow down local builds (in which we rely on `--mount-type=cache`, but now any change in go.sum will cause full redownload). I'm not sure how to solve this so it works nicely both in CI and locally though :(

kwapik: This does not work anyway, co closing this PR.

krdln: Trying a different approach, hope it works https://github.com/fluxninja/aperture/pull/2149

PR #2180, opened 2023-06-21
Added metrics labels for flux meter and telemetry collectors

hdkshingala:
##### Checklist

### Summary by OpenAI

**New Feature:**
- Added `policy_name` field to FluxMeterWrapper, TelemetryCollectorWrapper, and InfraMeterWrapper messages
- Updated PromQL queries and dashboards to include `policy_name` label
- Modified OpenTelemetry configuration to handle new InfraMeterWrapper type

**Refactor:**
- Changed `customConfig` map to `allInfraMeters` map in otel-config.go
- Updated AddInfraMeters function to accept new parameter types

> 🎉 A new dawn for policies, we cheer! 🥳
> With names attached, the path is clear. 🛣️
> Dashboards refined, metrics aligned, 📊
> In this PR, great progress we find! 🌟


pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -31,7 +34,28 @@ func AddInfraMeters(
...
+					"action": "upsert",
+					"value":  policyName,
+				},
+				{
+					"key":    "aperture_telemetry_collector_id",
+					"action": "upsert",
+					"value":  telemetryCollectorID,
+				},
+			},
+		})
```
krdln, line 57: Please move adding this processor to `addInfraMeter`

hdkshingala: @krdln I have added it there earlier but we had to pass all those variables from current function to the `addInfraMeter` function so I kept it here. Are you suggesting to add this processor totally somewhere else?

krdln: You mean these vars?

> 
> ```
> 	policyName string,
> 	telemetryCollectorID int64,
> ```

Yeah, I think it's cleaner to pass these to `addInfraMeter`. Before all the logic related to the single infra meter was in `addInfraMeter`, but right now a processor is created in `AddInfraMeters` and rest of pipeline in `addInfraMeter` which was confusing to me on the first look

cmd/aperture-agent/agent/otel-config.go
```diff
@@ -89,7 +86,7 @@ func provideAgent(
 				ims[fmt.Sprintf("%s/%s", prefix, k)] = v
 			}
 		}
-		if err := inframeter.AddInfraMeters(otelCfg, ims); err != nil {
+		if err := inframeter.AddInfraMeters(otelCfg, tc.GetPolicyName(), tc.GetTelemetryCollectorId(), ims); err != nil {
```
krdln, line 89: This is now incorrect:

```
		otelCfg := baseOtelCfg.MustCopy()
		ims := map[string]*policylangv1.InfraMeter{}
		for prefix, v := range allInfraMeters {
			for k, v := range v {
				ims[fmt.Sprintf("%s/%s", prefix, k)] = v
			}
		}
		if err := inframeter.AddInfraMeters(otelCfg, tc.GetPolicyName(), tc.GetTelemetryCollectorId(), ims); err != nil {
			log.Error().Err(err).Msg("unable to add custom metrics pipelines")
			utils.Shutdown(shutdowner)
			return
		}
```

`ims` contain all inframeters – not only from current policy.

krdln (changes requested):

cmd/aperture-agent/agent/otel-config.go
hdkshingala: Added a fix for this

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: Made the changes. Thank you!

cmd/aperture-agent/agent/otel-config.go
```diff
@@ -62,34 +59,37 @@ func provideAgent(
...
-		for prefix, v := range allInfraMeters {
-			for k, v := range v {
-				ims[fmt.Sprintf("%s/%s", prefix, k)] = v
-			}
-		}
-		if err := inframeter.AddInfraMeters(otelCfg, ims); err != nil {
+		if err := inframeter.AddInfraMeters(otelCfg, tc.GetPolicyName(), tc.GetTelemetryCollectorId(), allInfraMeters); err != nil {
 			log.Error().Err(err).Msg("unable to add custom metrics pipelines")
 			utils.Shutdown(shutdowner)
 			return
```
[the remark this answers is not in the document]
hdkshingala: This is a read operation only so lock is not required here

cmd/aperture-agent/agent/otel-config.go
```diff
@@ -62,38 +59,44 @@ func provideAgent(
...
+		if err := inframeter.AddInfraMeters(otelCfg, allInfraMeters); err != nil {
+			allInfraMetersMutex.RUnlock()
 			log.Error().Err(err).Msg("unable to add custom metrics pipelines")
 			utils.Shutdown(shutdowner)
 			return
 		}
+		allInfraMetersMutex.RUnlock()
 		// trigger update
 		log.Info().Msgf("received infra meter update, hot re-loading OTel, total infra meters: %d", len(allInfraMeters))
 		configProvider.UpdateConfig(otelCfg)
```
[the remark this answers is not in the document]
hdkshingala: Your suggested change is what is already added here

cmd/aperture-agent/agent/otel-config.go
```diff
@@ -62,34 +59,35 @@ func provideAgent(
...
-		for prefix, v := range allInfraMeters {
-			for k, v := range v {
-				ims[fmt.Sprintf("%s/%s", prefix, k)] = v
-			}
-		}
-		if err := inframeter.AddInfraMeters(otelCfg, ims); err != nil {
+		if err := inframeter.AddInfraMeters(otelCfg, allInfraMeters); err != nil {
 			log.Error().Err(err).Msg("unable to add custom metrics pipelines")
 			utils.Shutdown(shutdowner)
 			return
```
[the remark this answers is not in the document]
hdkshingala: We have locked the entire block where the allInfraMeters is used so there are no chances that the field would be accessed concurrecntly

pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -28,22 +30,48 @@ func AddInfraMeters(
...
+	policyName string,
+	telemetryCollectorID int64,
+	telemetryCollectorName string,
+	infraMeterName string,
 	infraMeter *policylangv1.InfraMeter,
 ) error {
-	pipelineName = strings.TrimPrefix(pipelineName, "metrics/")
+	processorName := fmt.Sprintf("%s-%s-%s-%d", otelconsts.ProcessorTelemetryCollector, policyName, infraMeterName, telemetryCollectorID)
+	config.AddProcessor(processorName, map[string]any{
+		"attributes": []map[string]interface{}{
```
krdln, line 56: Nit: you can just add this processor to inframeter.Processors & Pipeline, then you won't need to invent a name for this processor.

(Note that this would need some care to not break the "implicit pipeline" logic.

krdln (approved):

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: As discussed, we will handle it in a follow up PR as we have to test the updated logic properly.

PR #2199, opened 2023-06-27
Add experimental support for managing multiple controllers

kklimonda-fn:
##### Checklist

- [ ] Breaking changes

### Summary by CodeRabbit

**New Feature:**
- Introduced support for managing multiple controllers in a Kubernetes cluster.
- Added optional TLS support for the controller deployment.
- Implemented dynamic naming for ConfigMap, Deployment, and ServiceAccount based on the instance.

**Refactor:**
- Updated import statements across various modules for better modularity.
- Enhanced error handling for unknown keys in the `params` object.
- Improved configuration merging logic.

**Bug Fix:**
- Fixed duplicate entries in changesets.

> 🎉 Here's to the code that now gleams,
> With dynamic names and multiple streams.
> TLS secure, no more static lure,
> Our Kubernetes controllers are now mature! 🚀


operator/controllers/controller/reconciler.go
```diff
@@ -366,10 +382,19 @@ func (r *ControllerReconciler) manageResources(ctx context.Context, log logr.Log
...
+	}
+
+	return fmt.Sprintf("controller-%s", instance.GetName())
+}
+
 // reconcileConfigMap prepares the desired states for Controller configmaps and
 // sends an request to Kubernetes API to move the actual state to the prepared desired state.
 func (r *ControllerReconciler) reconcileConfigMap(ctx context.Context, instance *controllerv1alpha1.Controller) error {
-	configMap, err := configMapForControllerConfig(instance.DeepCopy(), r.Scheme)
+	name := r.createNameForConfigMap(instance)
```
hdkshingala, line 396: With the flag, this should be fine as we are using the older names here.

The second option is, we can handle the migration to new names here as below:
1. Use the names based on the instance
2. In each of the reconcile functions like `reconcileConfigMap`, we can first create/update the new resources based on the names from first point.
3. We can then do a check if the Kubernetes resources with previous name `aperture-controller` exists.
4. If it exists, we can delete them
5. Use status to set that the clean-up of older resources is done so we don't repeat points 3-4 on each reconcilation loop.
6. We can then add a deprecation message that this is only supported in next 3 releases (like we do for other areas) and then remove this clean-up logic.

User can still change this field as a trial but we can assume that they won't dig that dipper while upgrading and merge this PR for now. We can handle the clean-up part in a follow-up PR

hdkshingala (approved):

operator/controllers/controller/reconciler.go
kklimonda-fn: Instead of doing that on per-resource basis, I've added deleteSingletonResources function that handles all of them at the same time. I've not added the spec change, but the reconciliation loop shouldn't run too often for that to become a problem?

operator/main.go
```diff
@@ -74,6 +75,8 @@ func main() {
 	flag.BoolVar(&controllerManager, "controller", false,
 		"Enable manager for Aperture Controller. "+
 			"Enabling this will ensure that Controller Custom Resource is monitored by the Operator.")
+	flag.BoolVar(&multipleControllersEnabled, "experimental-multiple-controllers", false,
```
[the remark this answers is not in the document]
kklimonda-fn: chart actually does expose args that we can use for that, although probably not in the best manner, as we also have to make sure `--leader-election=True` is still on the list:

```
        args+: [
          "--experimental-multiple-controllers=True",
          "--leader-elect=True"
        ],
```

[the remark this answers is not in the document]
hdkshingala: Yes, right. We add `--leader-elect=True` by default from template so we just have to add `--experimental-multiple-controllers=True`

operator/controllers/controller/reconciler.go
```diff
@@ -248,6 +268,55 @@ func (r *ControllerReconciler) updateStatus(ctx context.Context, instance *contr
 	return nil
 }
 
+func (r *ControllerReconciler) deleteSingletonResources(ctx context.Context, log logr.Logger, req ctrl.Request, instance *controllerv1alpha1.Controller) error {
+	// If we created controller with name "controller" its full name will match singleton name.
+	instanceFullName := fmt.Sprintf("%s-%s", controllers.AppName, instance.GetName())
+	if instanceFullName == controllers.ControllerServiceName {
```
hdkshingala, line 274: This should be `!=`, right? As we are trying to do deletion for resources with older names after this change? Or I misunderstood this function?

operator/controllers/controller/reconciler.go
hdkshingala: Yes we have handled the reconcile loop so that it is not triggered too frequently on start-up (it hits 4-5 calls in a second otherwise) so it should be fine

operator/controllers/controller/reconciler.go
```diff
@@ -64,6 +65,29 @@ var (
 	controllerClientCert *bytes.Buffer
 )
 
+func (r *ControllerReconciler) verifySingletonControllerExists(ctx context.Context, instance *controllerv1alpha1.Controller, instances *controllerv1alpha1.ControllerList) (bool, error) {
+	// Check if this is an update request. If not, skip the resource creation as only single Controller is required in a cluster.
+	if instances.Items != nil && len(instances.Items) != 0 {
```
IridiumOxide, line 70: len(nil) is 0, this condition can be simplified by removing the first part

IridiumOxide: The condition can be dropped altogether as `range` supports nil slices

IridiumOxide (approved):

operator/controllers/controller/reconciler.go
kklimonda-fn: We want to avoid deletion of the resource in case instanceFullName matches controllers.ControllerServiceName, otherwise reconcile loop will constantly delete and recreate it.

Issue #2212, opened 2023-06-28
Use vtprotobuf for faster unmarshalling and less mallocs


krdln:
Links:
* https://vitess.io/blog/2021-06-03-a-new-protobuf-generator-for-go/
* https://buf.build/community/planetscale-vtprotobuf

Performance improvements should come from two aspects:
* vtprotobuf avoids reflect and generate specialized code,
* vtprotobuf allows pooling object to reduce allocs.

krdln:
Discovered that pooling in vtprotobuf is underpowered (this is known – there are issues on the repo):
* doesn't work for oneofs,
* doesn't work for slices of messages,
* doesn't work for maps.

Basically, in practice pooling basically works for preserving capacity for slices of strings, which _is_ something, but leaves lot of allocs on the table. Also, note that the major source of allocs in proto unmarshalling are strings, which is something not fixable by pools in Go.

PR #2213, opened 2023-06-28
perf: Speed up proto unmarshaling using vtprotobuf plugin

krdln:
### Description of change

Speed up marshaling / unmarshaling of check(http) requests & responses by using
optimized methods generated by vtprotobuf plugin and some use of pooling.

Results: ~6% speedup (total (un)marshaling: ~15% → ~8%), 10% reduction in allocs.

Drive-by: Expose heap profiler endpoint.

Resolves https://github.com/fluxninja/aperture/issues/2212

Possible followup: Would be nice to limit the vtprotobuf code generation to only check apis.

##### Checklist


krdln:
obraz

PR #2224, opened 2023-06-29
Added kubeletstats infra_meter in auto_scaling and load_scheduling blueprints

hdkshingala:
- Added all the common configurations of Policy and Dashboards under `common/config-defaults.libsonnet`
- Removed the duplicate config from all the blueprints
- Added `kubeletstats_inframeter` under auto_scaling blueprint with enable=true by default
- Added `kubeletstats_inframeter` under load_scheduling blueprint with enable=false by default. Reason for adding it in defaults is, we have overload_confirmation for `adaptive_load_scheduler` so if someone needs it, they just have to change a single field.
- Added `kubeletstats_inframeter` under postgresql blueprint variant with enable=true by default

### Summary by CodeRabbit

**New Features:**
- Introduced `kubeletstats_infra_meter` configuration to auto-scaling and load-scheduling blueprints.
- Refactored common configurations into a shared file for better maintainability.
- Modified the structure of policy and dashboard objects.
- Added new controllers for pod CPU and memory scaling in the auto-scaling blueprint.

**Refactor:**
- Updated logic of `policyDef` function to include `kubeletstats_inframeter` in the resources of the policy.
- Removed duplicate configurations and centralized them in a common configuration file.

**Documentation:**
- Updated documentation to reflect changes in parameters, their order, and descriptions.

> 🎉 A refactor so neat, a feature so grand,  
> In the realm of code, we make our stand.  
> With `kubeletstats_infra_meter` in hand,  
> Scaling pods on demand, just as planned! 🚀


blueprints/auto-scaling/common/config-defaults.libsonnet
```diff
@@ -21,6 +9,25 @@ local scaling_parameters_defaults = {
   },
 };
 
+/**
+* @schema (kubeletstats_infra_meter_label_filter.key: string) Key represents the key or name of the field or labels that a filter can apply on.
```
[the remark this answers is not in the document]
hdkshingala: Sure, will make changes as per the config suggestions.

blueprints/auto-scaling/pod-auto-scaler/config.libsonnet
```diff
@@ -1,9 +1,34 @@
 local autoScalingDefaults = import '../common/config-defaults.libsonnet';
 
+/**
+* @schema (scaling_driver.scale_out.enabled: bool) Enables the driver to do scale out of the resource.
```
tanveergill, line 4: Seems to be mismatch between schema and defaults below: `scaling_driver` vs `scaling_threshold` at https://github.com/fluxninja/aperture/pull/2224/files?diff=split&w=0#diff-531ae09bddf21b0099d51bed37dcb3b59003edc91f1977d5c949e1e3976faa5bR33

hdkshingala: Oh yes, I renamed the field later but missed to update here. Will update it

PR #2355, opened 2023-07-11
De-duplicated receivers and processors of the infra meter OTEL Pipelines

hdkshingala:
image

Note: Brings in multi-selectors in k8s attribute processor https://github.com/fluxninja/opentelemetry-collector-contrib/commit/6f18e2a71b530ed62447dd84310392d44d65889a

### Summary by CodeRabbit

**Refactor:**
- Modified the `kubeletstats_infra_meter` function to replace the `filters` parameter with `selectors`, and added a check to ensure that `selectors` is an array.
- Updated the value of `ProcessorK8sAttributes` from `"k8sattributes/kubeletstats"` to `"k8sattributes"`.
- Introduced a new constant `ProcessorK8sAttributesSelectors` with the value `"selectors"`.

**New Feature:**
- Enhanced the `addInfraMeter` function to handle duplication of receivers and processors in the infra meter OTEL pipelines using SHA-256 hashing for unique identification.
- Added logic to de-duplicate receivers and processors in the infra meter OpenTelemetry pipelines, handling per-agent-group conditions, and updating Kubernetes attribute processors with new selectors.

**Bug fix:**
- Improved error message in the `otelState` function to include the current state of the otel-collector when it's not running.

> 🎉 Here's to the code that's now more robust, 🥂
> With selectors in place, and filters lost. 🗑️
> Duplication handled, no more fuss, 🚀
> And better error messages, at no extra cost! 💪


pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -70,19 +73,88 @@ func addInfraMeter(
...
+			return fmt.Errorf("failed to prepare sha256 for receiver '%s' in infra_meter: '%s' of Policy: '%s': %w", origName, infraMeterName, policyName, err)
+		}
+		id = component.NewIDWithName(id.Type(), fmt.Sprintf("%x", configSHA.Sum(nil)))
+		strID := id.String()
+
+		// If receiver is already present with given id and per-agent-group = false, skip adding receiver with per-agent-group = true.
+		if _, ok := config.Receivers[strID]; ok && infraMeter.PerAgentGroup {
+			receiverIDs[origName] = strID
+			continue
+		}
```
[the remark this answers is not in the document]
hdkshingala: That is handled at line no 96

pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -70,19 +73,88 @@ func addInfraMeter(
...
+			selectors, ok = processorConfig.Fields[otelconsts.ProcessorK8sAttributesSelectors]
+			if ok && selectors != nil {
+				selectorsList = selectors.GetListValue().AsSlice()
+				delete(processorConfig.Fields, otelconsts.ProcessorK8sAttributesSelectors)
+			}
+		}
+
+		configSHA := sha256.New()
+		if _, err := configSHA.Write([]byte(processorConfig.String())); err != nil {
+			return fmt.Errorf("failed to prepare sha256 for receiver '%s' in infra_meter: '%s' of Policy: '%s': %w", origName, infraMeterName, policyName, err)
```
krdln, line 138: sha.Write never fails. You can safely panic / ignore error

pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -143,13 +214,24 @@ func mapSlice(mapping map[string]string, xs []string) []string {
...
 }
 
-// normalizeComponentName normalizes user defines component name by adding
-// `user-defined-<pipeline_name>` suffix.
-// This ensures no builtin components are overwritten.
-func normalizeComponentName(pipelineName, componentName string) string {
-	suffix := fmt.Sprintf("user-defined-%s", pipelineName)
-	if componentName == "" {
-		return suffix
+func updateK8sAttributesProcessor(processor interface{}, selectorsList []interface{}) map[string]interface{} {
```
krdln, line 217: Nit: no need to return a new map, you're modifying it in place anyway

krdln (approved):
lgtm

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: golangci-lint was complaining about it so I kept it.

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: But we don't take the map as input and it is initialised in the function only

krdln: we take the processor from `config.Processors[strID]`, from which we get the map. which we edit in place, right?

hdkshingala: We get interface from config.Processors[strID] and then type cast it to map

pkg/otelcollector/infra-meter/infra-meter.go
krdln: btw, sha256 defines helper function https://pkg.go.dev/crypto/sha256#Sum256, this should simplify

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: Made the changes. Thank you!

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: Made the changes using their helper

krdln (approved):

PR #2358, opened 2023-07-11
fix: remove injectDecoder after controller-runtime update to v0.15.0

slayer321:
##### Checklist

- [ ] Breaking changes

### Summary by CodeRabbit

```markdown
### Refactor:
- Updated `ApertureInjector` struct and its `Handle` method in `mutating_webhook_handler.go`, making the `decoder` field public and removing the `InjectDecoder` method.
- Introduced a new `decoder` object from the `admission` package of `controller-runtime` in `main.go`.
- Removed the `decoder` field and the `InjectDecoder` method from the `AgentHooks` and `ControllerHooks` structs in `hook.go` files.

These changes are due to an update in the `controller-runtime` library, improving the code's maintainability and efficiency.
```

> 🎉 Here's to the code that we've refactored,  
> To the runtime library, our thanks is offered.  
> With every change, our code gets smarter,  
> On the path of progress, we march farther! 🚀


operator/main.go
```diff
@@ -124,6 +125,7 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	decoder := admission.NewDecoder(mgr.GetScheme())
```
hdkshingala, line 128: We don't need a variable for this as it is only used once

hdkshingala (approved):

Issue #2196, opened 2023-06-26
Rename signals to be more descriptive


DariaKunoichi:
**Is your feature request related to a problem? Please describe.**
With new dashboard generator there is a generic way to add a panel for simple promql query - it shows a panel title like "Query for <SIGNAL>"

**Describe the solution you'd like**
Replace signals names in policies to be more descriptive.

PR #2359, opened 2023-07-11
Rename signals to be more descriptive

DariaKunoichi:
### Description of change
Closes: GH-2196

##### Checklist

- [ ] Breaking changes

### Summary by CodeRabbit

**Refactor:**
- Renamed signal names from 'SIGNAL' to 'LATENCY' and 'SIGNAL_LONG_TERM' to 'LATENCY_LONG_TERM' across multiple files for better clarity and understanding.

> 🎉 With a name change, clear as day, 🌞
> 
> 'SIGNAL' to 'LATENCY', we pave the way. 🛣️
> 
> 'LONG_TERM' too joins the fray, 🏁
> 
> Making our code more intuitive, hooray! 🥳


blueprints/load-scheduling/common/policy.libsonnet
```diff
@@ -83,7 +83,7 @@ function(cfg, params={}, metadata={}) {
       + spec.v1.AdaptiveLoadScheduler.withDryRun(updatedConfig.policy.service_protection_core.dry_run)
       + spec.v1.AdaptiveLoadScheduler.withInPorts({
         overload_confirmation: (if isConfirmationCriteria then spec.v1.Port.withSignalName('OVERLOAD_CONFIRMATION') else spec.v1.Port.withConstantSignal(1)),
-        signal: spec.v1.Port.withSignalName('SIGNAL'),
```
tanveergill, line 86: This is going to break the other variants since they may be based on an arbitrary signal from promql? Instead change the signature of `function` in the common code to take the names of signal and setpoint signals so that they may be customized according to the variant.

DariaKunoichi: you are right, thanks for noticing

PR #2378, opened 2023-07-13
Fix OTel ConfiProvider

hasit:
##### Checklist

- [ ] Breaking changes

### Summary by CodeRabbit

**Release Notes:**

- **Refactor**: Introduced new locks in `Provider` struct for concurrent access protection to `config`, `watchFunc`, and `hooks` fields.
- **New Feature**: Added a new route handler for `/goroutine` in the `setupProfilers` function for goroutine profiling.
- **Test**: Updated `otel_test.go` to use `GetConfig` method instead of `MustGetConfig` from `configProvider` object, improving error handling.

> 🎉 With locks that guard and routes anew,
> 
> Tests refined, their purpose true.
> 
> In code we trust, in change we thrive,
> 
> To the future of our codebase, we high-five! 🚀

### DIFF
--- a/pkg/otelcollector/config/provider.go
+++ b/pkg/otelcollector/config/provider.go
@@ -5,8 +5,9 @@ import (
 	"errors"
 	"sync"
 
-	"github.com/fluxninja/aperture/v2/pkg/log"
 	"go.opentelemetry.io/collector/confmap"
+
+	"github.com/fluxninja/aperture/v2/pkg/log"
 )
 
 // comply with confmap.Provider interface.
@@ -16,11 +17,12 @@ var _ confmap.Provider = (*Provider)(nil)
 //
 // It allows updating the config and registering hooks.
 type Provider struct {
-	lock      sync.Mutex // protects config, watchFunc & hooks
-	config    *Config    // nil only after Shutdown.
-	watchFunc confmap.WatcherFunc
-	hooks     []func(*Config)
-	scheme    string
+	configLock    sync.RWMutex // protects config, watchFunc & hooks
+	watchFuncLock sync.Mutex   // protects watchFunc
+	config        *Config      // nil only after Shutdown.
+	watchFunc     confmap.WatcherFunc
+	hooks         []func(*Config)
+	scheme        string
 }
 
 // NewProvider creates a new OTelConfigProvider.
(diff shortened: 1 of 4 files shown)


tanveergill (approved):

pkg/otelcollector/config/provider.go
```diff
@@ -103,26 +89,54 @@ func (p *Provider) setConfig(config *Config) {
...
+
+// GetConfig returns the current config.
+func (p *Provider) GetConfig() *Config {
+	p.configLock.RLock()
+	defer p.configLock.RUnlock()
+	return p.config
+}
+
+// SetConfig sets the new config, replacing the old one.
+func (p *Provider) SetConfig(config *Config) {
```
krdln, line 113: Why all the "internal" functions are public? This just begs for breaking invariants

pkg/otelcollector/config/provider.go
```diff
@@ -103,26 +89,54 @@ func (p *Provider) setConfig(config *Config) {
...
 func (p *Provider) AddMutatingHook(hook func(*Config)) {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-
 	if p.config == nil {
 		log.Warn().Msg("OtelConfigProvider.AddHook: already shut down")
 		return
 	}
 
 	p.hooks = append(p.hooks, hook)
```
krdln, line 100: p.hooks and p.config are not protected after these changes.

pkg/otelcollector/config/provider.go
```diff
@@ -103,26 +89,54 @@ func (p *Provider) setConfig(config *Config) {
...
-	p.setConfig(p.config)
+	p.UpdateConfig(p.config)
+}
+
+// GetConfig returns the current config.
+func (p *Provider) GetConfig() *Config {
+	p.configLock.RLock()
+	defer p.configLock.RUnlock()
+	return p.config
+}
```
krdln, line 110: Why not return copy? Are we sure noone will edit the returned config?

pkg/otelcollector/config/provider.go
```diff
@@ -103,26 +89,54 @@ func (p *Provider) setConfig(config *Config) {
...
+		log.Warn().Msg("OtelConfigProvider: tried to update config after Shutdown")
+		return
+	}
+	p.config = config
+}
+
+// GetWatchFunc returns the current watch function.
+func (p *Provider) GetWatchFunc() confmap.WatcherFunc {
+	p.watchFuncLock.Lock()
+	defer p.watchFuncLock.Unlock()
```
krdln, line 133: what's the rationale for locking only in getters instead of top-level functions? This makes it easier to break invariants

Issue #2402, opened 2023-07-20
Allow providing api key to aperturectl via env or file


krdln:
Initial proposal:
* ~~Add `APERTURECTL_API_KEY` that stores _path_ to the file which contains controller's api key.~~
* If that flag is present, we change the default of connecting to `--kube` to connect to production cloud controller using that api key (no `--controller` flag is necessary).
* We also need some kind of way to use that env variable with different address of controller than production (for internal testing). We don't need to document it though.

krdln:
This is my understanding of what we've agreed on:

Config file (toml or ini or json whatever you choose. I like toml, but I'm ok with others), something like:

```toml
[controller]
api-key = "xxx" // perhaps call it `token`, not api key? Lot of places call it token, eg. github, crates.io, etc.
address = "org.fluxninja.com" // or perhaps just organization and we can create ctrl addr from it?
```

Config file default location: `~/.aperturectl/config` or `~/.aperturectl/config.toml` – not sure if we should include extension or not :shrug: 

Config file env variable `APERTURECTL_CONFIG`

Config file flag for aperturectl `--config CONFIG_PATH`.

Precedence:
1. command line (`--kube`, `--controller`, `--api-key`)
2. Config from `--config PATH`
3. Config from env
4. Config from default location

Note: configs don't merge, we load one config.

If there's no config file, we default to `--kube`.

Optionally we can implement this opentelemetry style syntax for interpolating env variables which would work for cmdline args, although I think that with config files, that's too much possible solutions.

PR #2403, opened 2023-07-20
Added support to load config from environment variables in aperturectl

hdkshingala:
Ref #2402

##### Checklist

### Summary by CodeRabbit

Release Notes:

- Documentation: Added support to load config from environment variables in `aperturectl`. The documentation now reflects the addition of this feature across various commands, enhancing the flexibility of the tool.

> "Config from env vars, a powerful feat,
> Aperturectl now more flexible and neat.
> With docs updated, users can see,
> How to configure with simplicity."

### DIFF
--- a/cmd/aperturectl/cmd/utils/controller.go
+++ b/cmd/aperturectl/cmd/utils/controller.go
@@ -13,7 +13,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	toml "github.com/pelletier/go-toml"
+	toml "github.com/BurntSushi/toml"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"google.golang.org/grpc"
@@ -33,6 +33,17 @@ import (
 )
 
 var controllerNs string
+
+// ControllerConfig is the config file structure for ARC Controller.
+type ControllerConfig struct {
+	URL    string `toml:"url"`
+	APIKey string `toml:"api_key"`
+}
+
+// Config is the config file structure for Aperture.
+type Config struct {
+	Controller *ControllerConfig `toml:"controller"`
+}
 
 // ControllerConn manages flags for connecting to controller – either via
 // address or kubeconfig.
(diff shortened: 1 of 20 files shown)


cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -115,6 +145,43 @@ func (c *ControllerConn) PreRunE(_ *cobra.Command, _ []string) error {
...
+		config, err := toml.LoadFile(updatedConfig)
+		if err != nil {
+			return fmt.Errorf("failed to read config file '%s': %w", updatedConfig, err)
+		}
+
+		if config == nil {
+			return fmt.Errorf("invalid config file '%s'", updatedConfig)
+		}
+
+		controller, ok := config.Get("controller").(*toml.Tree)
```
krdln, line 158: Please unmarshal config to a struct.

cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -9,8 +9,11 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
 
+	toml "github.com/pelletier/go-toml"
```
krdln, line 16: Consider using https://github.com/BurntSushi/toml if possible – I see we already have it in go.mod and it has more starts on gh.

krdln (commented):
Please un config to struct

cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -92,15 +96,41 @@ func (c *ControllerConn) InitFlags(flags *flag.FlagSet) {
...
+			if _, err := os.Stat(env); err != nil {
+				env = ""
+			}
+		}
+	}
+
+	updatedConfig := c.config
+	if updatedConfig == "" {
+		updatedConfig = env
+	}
```
krdln, line 123: I think this way env overrides commandline flag, it should be the other way.

cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -92,15 +96,41 @@ func (c *ControllerConn) InitFlags(flags *flag.FlagSet) {
...
+		homeDir, err := os.UserHomeDir()
+		if err == nil {
+			env = filepath.Join(homeDir, ".aperturectl", "config")
+			if _, err := os.Stat(env); err != nil {
+				env = ""
+			}
+		}
+	}
+
+	updatedConfig := c.config
```
krdln, line 120: Nit: The `updatedConfig` name is very confusing. Can we just avoid using different variable and use `c.config`?

If you want to keep this check
```
	if c.config != "" && c.apiKey != "" {
		return errors.New("--api-key cannot be used with --config")
	}
```

it can be moved above.

Or perhaps, move all the config-path resolution down to line 111 (just before `if c.kube`). That way, flag verification will happen before we check env variables, which make sense to me.

cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -92,15 +107,38 @@ func (c *ControllerConn) InitFlags(flags *flag.FlagSet) {
...
+	}
+
+	if c.controllerAddr == "" && !c.kube && c.config == "" {
+		log.Info().Msg("Neither --controller nor --kube nor --config flag is set. Assuming --kube=true.")
 		c.kube = true
 	}
 
+	if c.config != "" && c.apiKey != "" {
+		return errors.New("--api-key cannot be used with --config")
+	}
```
krdln, line 140: This check should be moved up, because it's ok if we use --api-key to override default config or config from env. (Actually, I think it's also ok if we have `--config xxx --api-key yyy and we override api key, but perhaps let's error, as that's weird)

cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -115,6 +153,33 @@ func (c *ControllerConn) PreRunE(_ *cobra.Command, _ []string) error {
 		if err != nil {
 			return err
 		}
+	} else if c.apiKey == "" && !c.kube {
```
krdln, line 156: Should it be just `else if c.config != ""`?

Otherwise if someone has local controller without api keys (but not via --kube), we will try to load config.

krdln (changes requested):

cmd/aperturectl/cmd/utils/controller.go
hdkshingala: Made the changes as suggested

cmd/aperturectl/cmd/utils/controller.go
hdkshingala: Made the changes. Thank you

PR #2421, opened 2023-07-25
docs: Showcase aperturectl with cloud controller in Get Started

krdln:
### Description of change

* Show how to configure aperturectl
* Show how to list and delete policies with aperturectl

##### Checklist

## Release Notes

- **Documentation**: Updated and added new guides for configuring `aperturectl` with a FluxNinja ARC endpoint and using it with the cloud controller. The instructions cover policy management and visualization of policy runtime.
- **Bug fix**: Corrected the spelling of "aperturectl" in the documentation.
- **New Feature**: Removed the requirement for `api_key` in the configuration file for self-hosted setups.

> 🎉 Changes are here, no need to fear,
> Documentation clear, for all to hear.
> No more typo smear, that's a cheer!
> Self-hosted near, let's give a cheer! 🥳

### DIFF
--- a/cmd/aperturectl/cmd/utils/controller.go
+++ b/cmd/aperturectl/cmd/utils/controller.go
@@ -174,6 +174,10 @@ func (c *ControllerConn) PreRunE(_ *cobra.Command, _ []string) error {
 			return fmt.Errorf("invalid config file '%s'. Missing key 'controller.url'", c.config)
 		}
 
+		if config.Controller.APIKey == "" {
+			return fmt.Errorf("invalid config file '%s'. Missing key 'controller.api_key'", c.config)
+		}
+
 		c.controllerAddr = config.Controller.URL
 		c.apiKey = config.Controller.APIKey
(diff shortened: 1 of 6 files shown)


cmd/aperturectl/cmd/utils/controller.go
```diff
@@ -174,10 +174,6 @@ func (c *ControllerConn) PreRunE(_ *cobra.Command, _ []string) error {
 			return fmt.Errorf("invalid config file '%s'. Missing key 'controller.url'", c.config)
 		}
 
-		if config.Controller.APIKey == "" {
```
hdkshingala, line 177: Are we removing this due to some reason? I think it is a mandatory field for cloud controller communication?

krdln: Yes, but realized that config files can be utilized also for self-hosted controller, so it's ok to not have api key. Feel free to revert it if you disagree

PR #2422, opened 2023-07-27
Circuit Viz: Display the original config of components using nested circuit

tanveergill:
##### Checklist

### Release Notes

**Refactor:**
- Introduced a new function `NewNestedCircuitAndOptions` for creating nested circuits and their options, replacing the previous `ParseNestedCircuit` function.
- Refactored various component parsing functions to return both a `runtime.ConfiguredComponent` and a `policylangv1.NestedCircuit`.
- Removed the `name` and `short_description` fields from the `NestedCircuit` message in the protobuf file.

**New Feature:**
- Added support for arbitrary configuration of nested circuit components using `google/protobuf/any.proto`.
- Included original configuration of components in nested circuits for better observability.

**Chore:**
- Updated documentation to reflect changes in the nested circuit component configuration.

> 🎉 Changes abound, code refactored round,  
> Nested circuits now with config found.  
> No more name or short description,  
> Just pure functionality in this rendition! 🚀


pkg/policies/controlplane/components/flowcontrol/load-scheduler/adaptive-load-scheduler.go
```diff
@@ -69,9 +70,15 @@ func ParseAdaptiveLoadScheduler(
...
+		return nil, err
+	}
+
 	nestedCircuit := &policylangv1.NestedCircuit{
 		Name:             "AdaptiveLoadScheduler",
 		ShortDescription: iface.GetSelectorsShortDescription(adaptiveLoadScheduler.Parameters.LoadScheduler.GetSelectors()),
+		Config:           config,
 		InPortsMap:       nestedInPortsMap,
 		OutPortsMap:      nestedOutPortsMap,
 		Components: []*policylangv1.Component{
```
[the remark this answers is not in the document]
tanveergill: okay

api/aperture/policy/language/v1/policy.proto
```diff
@@ -357,4 +358,6 @@ message NestedCircuit {
   string name = 4;
   // Short description of the nested circuit component. This description is displayed by graph visualization tools.
   string short_description = 5;
+  // Arbitrary configuration for the nested circuit component. This configuration is displayed by graph visualization tools.
+  google.protobuf.Any config = 6;
```
harjotgill, line 362: This will show up in the spec now, right? Should we just provide this as a string or something?

tanveergill: Yes, do you see a problem with that? I don’t really need to do this through the spec, can set the config in code instead

tanveergill: Name and description fields were already there so added this. But we can very much set these fields in code where required instead of relying on the spec

tanveergill: Pushed a new commit, taking code-based approach now. Removed all visualization specific fields from the policy spec.

tanveergill:
Screenshot 2023-07-28 at 2 06 23 PM

PR #2478, opened 2023-08-07
Elasticseach setup and client to generate the load

hdkshingala:
### Summary by CodeRabbit

**Release Notes**

- **New Feature**: Integrated Elasticsearch into the application, enabling advanced search capabilities and improved data handling.
- **Refactor**: Simplified logic in `add_kubelet_overload_confirmations` function for better readability and maintainability.
- **New Feature**: Introduced a new policy for load scheduling in Elasticsearch based on the search thread pool queue size metric.
- **Documentation**: Updated descriptions of parameters in various files to reflect accurate functionality.

> 🎉 With Elasticsearch now in sight,
> Our app takes a powerful flight. 
> Data handling's a breeze,
> And with policies that please,
> We're set for an incredible height! 🚀


blueprints/load-scheduling/common/utils.libsonnet
```diff
@@ -39,13 +39,7 @@ local add_kubelet_overload_confirmations(c) = {
   local updated_cfg = c {
     policy+: {
       service_protection_core+: {
-        local overloadConfirmations = if std.objectHas(c.policy.service_protection_core, 'overload_confirmations') then
```
hdkshingala, line 42: Why did we remove this? This was added for users who wants to use some other signals than CPU and memory usage and have provided the same via the values file. With this change, even if user provides some values under `overload_confirmations`, that would get skipped

hasit: With this line, `overload_confirmations` where being added twice, in turn, causing the generated policy to have double the amount of components.

harjotgill (approved):

PR #2482, opened 2023-08-08
docs: Fix links to Aperture Controller section

krdln:
This section is directly linked from some other places in docs, and our link checker doesn't dive into url fragments.

### Release Notes

**Documentation**: Improved navigation in the architecture documentation by adding an anchor tag to the "Aperture Controller" section. This change enhances user experience by allowing direct linking to this specific section.

> 📚 In the realm of code, where logic intertwines,
> 
> A tiny change emerges, as a beacon it shines.
> 
> To the Aperture Controller, a path we add,
> 
> Making navigation smoother, making users glad. 😊


DariaKunoichi (approved):
ok

PR #2495, opened 2023-08-11
Add logger for http server ErrorLogs

DariaKunoichi:
##### Checklist

- [ ] Breaking changes

### Summary by CodeRabbit

# Release Notes

### New Feature
- Introduced `apertureLogger`, a new type that implements the `io.Writer` interface. This feature redirects log output from the standard library logger to the global logger provided by the `log` package, ensuring all server errors are captured effectively.
- Updated `provideServer` function to use `apertureLogger` as the `ErrorLog` for the HTTP server, enhancing error logging capabilities.

### DIFF
--- a/pkg/net/http/server.go
+++ b/pkg/net/http/server.go
@@ -95,7 +95,7 @@ type apertureLogger struct {
 }
 
 func (al *apertureLogger) Write(p []byte) (n int, err error) {
-	al.logger.Error().Msg(string(p))
+	al.logger.Debug().Msg(string(p))
 	return len(p), nil
 }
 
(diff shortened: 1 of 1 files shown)


harjotgill (approved):

pkg/net/http/server.go
```diff
@@ -89,6 +90,15 @@ func (constructor ServerConstructor) Annotate() fx.Option {
...
 
+type apertureLogger struct {
+	logger *log.Logger
+}
+
+func (al *apertureLogger) Write(p []byte) (n int, err error) {
+	al.logger.Debug().Msg(string(p))
+	return len(p), nil
+}
+
```
harjotgill, line 101: Our existign logger already implements Write interface and can be passed directly.

Issue #2496, opened 2023-08-14
Bring back JMX blueprint & scenario


DariaKunoichi:

DariaKunoichi:
Waiting on overload_confirmations fixes.

PR #2509, opened 2023-08-17
Jmx policy

DariaKunoichi:
### Description of change
Closes: GH-2496

##### Checklist

- [ ] Breaking changes

obraz

### Summary by CodeRabbit

- New Feature: Added support for load scheduling based on JMX metrics. This includes new configurations, policies, and dashboards related to JMX metrics.
- New Feature: Introduced a function to generate Prometheus configuration for JMX metrics collection.
- New Feature: Added a new policy "Protection with JMX Overload Confirmation" and a corresponding dashboard variant.
- Refactor: Updated the function signatures in `policy.libsonnet` and `bundle.libsonnet` to include additional parameters.
- Test: Added a load test scenario in `playground/scenarios/jmx-service-protection/load-generator/test.js`.
- Documentation: Added a new documentation file `docs/content/reference/blueprints/load-scheduling/jmx.md` detailing the usage and configuration of the new features.


blueprints/load-scheduling/jmx/config.libsonnet
```diff
@@ -0,0 +1,47 @@
...
+      * @param (policy.latency_baseliner.long_term_query_interval: string) Interval for long-term latency query, i.e., how far back in time the query is run. The value should be a string representing the duration in seconds.
+      * @param (policy.latency_baseliner.long_term_query_periodic_interval: string) Periodic interval for long-term latency query, i.e., how often the query is run. The value should be a string representing the duration in seconds.
+      * @param (policy.latency_baseliner.latency_tolerance_multiplier: float64) Tolerance factor beyond which the service is considered to be in overloaded state. E.g. if the long-term average of latency is L and if the tolerance is T, then the service is considered to be in an overloaded state if the short-term average of latency is more than L*T.
+      */
+      long_term_query_interval: '1800s',
+      long_term_query_periodic_interval: '30s',
+      latency_tolerance_multiplier: 1.25,
+    },
+
+    jmx: {
```
hdkshingala, line 37: Are we using this somewhere in bundle/policy?

DariaKunoichi: we are using this in utils file that configures an infra-meter

hdkshingala: Ah yes, I see it now. Could we please move it to bundle like we do in other variants like - https://github.com/fluxninja/aperture/blob/main/blueprints/load-scheduling/postgresql/bundle.libsonnet#L20

The reason behind doing this in bundle is if the dashboard creator also wants to use it, it can and we don't have to duplicate this

DariaKunoichi: moved to bundle file as you instructed

hasit (approved):

tanveergill (changes requested):
@DariaKunoichi,

Here's the changes we need:

1. The approach we want to take with service protection variants is to use saturation metrics as the primary signals to adaptive load scheduler instead of latency. Derive from the PromQL blueprint that takes a signal and threshold. Let's do a single variant that works on the GC metric.
2. Current blueprint expects the query from the user. Instead, the default query needs be defined in the blueprint. It may be overridden by the user.

DariaKunoichi:
obraz

tanveergill (approved):

hdkshingala (approved):

DariaKunoichi:
obraz

Only one activation of request rejection - but it correlates with GC threshold of 20.

blueprints/load-scheduling/jmx/config.libsonnet
```diff
@@ -0,0 +1,28 @@
+local serviceProtectionDefaults = import '../common/config-defaults.libsonnet';
+
+
+serviceProtectionDefaults {
+  policy+: {
+    /**
+    * @param (policy.promql_query: string) PromQL query.
+    */
+    promql_query: 'avg(java_lang_G1_Young_Generation_LastGcInfo_duration{k8s_pod_name=~"service3-demo-app-.*"})',
```
tanveergill, line 9: This is a hardcoded query! No need to expose the query as we discussed. The blueprint needs to be specific for garbage collection. Filter by the infra meter name and policy name in the query

PR #2669, opened 2023-09-25
Fixes in blueprints

hdkshingala:
This PR contains following changes:

1. Fixed the hard-coded setpoint in Postgresql, Elasticsearch and JMX blueprints
2. Moved the config update logic to `bundle.libsonnet` as it would be needed for dashboard generation
3. Moved the `overload_condition` flag to root of component and fixed it based on use-case in blueprints

### Summary by CodeRabbit

- New Feature: Introduced a new configuration option `service_protection_core` with a `setpoint` field and an `overload_condition` field across multiple modules. This allows users to specify the criteria for determining overload state.
- Refactor: Simplified code by removing unnecessary updates of configuration objects and directly passing original configurations to functions.
- Documentation: Updated documentation to reflect changes in configuration parameters, including the addition of `service_protection_core.setpoint` and removal of `overload_condition` from various parameters.
- Chore: Adjusted internal handling of `overload_condition`, moving it from `Parameters` to top-level in `AIADLoadScheduler`.


blueprints/load-scheduling/promql/config.libsonnet
```diff
@@ -10,6 +10,12 @@ serviceProtectionDefaults {
     * @param (policy.setpoint: float64) Setpoint.
     */
     setpoint: '__REQUIRED_FIELD__',
+    /**
+    * @param (policy.service_protection_core.overload_condition: string)Overload condition determines the criteria to determine overload state. The condition must be one of: gt, lt, gte, lte.
+    */
+    service_protection_core+: {
+      overload_condition: '__REQUIRED_FIELD__',
```
tanveergill, line 17: This should be hardcoded for each recipe

hdkshingala: This is hard-coded in all of them except promql as query could be anything here

blueprints/load-scheduling/average-latency/policy.libsonnet
```diff
@@ -3,7 +3,13 @@ local commonPolicyFn = import '../common-aimd/policy.libsonnet';
 local config = import './config.libsonnet';
 
 function(cfg, params={}, metadata={}) {
-  local updatedConfig = config + cfg,
+  local updatedConfig = config + cfg + {
+    policy+: {
+      service_protection_core+: {
+        overload_condition: 'gt',
```
tanveergill, line 9: Shouldn’t be part of service protection core. This should be part of base blueprint just like PromQL query

blueprints/load-scheduling/promql/config.libsonnet
tanveergill: I see

PR #2896, opened 2023-11-02
Add support for secret manager in inframeter config update

hdkshingala:
##### Checklist

## Summary by CodeRabbit

New Features:
- Integrated the Secret Manager package into the Aperture Agent, enhancing security by managing sensitive data.
- Introduced a new struct "ProvideAgentIn" in the Aperture Agent configuration, improving code readability and maintainability.
- Expanded the functionality of the `AddInfraMeters` function in the Otel Collector package, allowing it to handle different installation modes and interact with the Secret Manager.

Refactor:
- Updated several function calls in the Aperture Agent configuration to use the new "ProvideAgentIn" struct, streamlining the code and improving clarity.
- Added new functions in the Otel Collector package to process secret references and fetch secret values, enhancing code organization and readability.

New Package:
- Added a new Secret Manager package that provides a client for managing secrets, improving security and data protection.


harjotgill (approved):

pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -231,3 +257,68 @@ func updateK8sAttributesProcessor(processor interface{}, selectorsList []interfa
...
+// processMapForSecrets processes the given map to search for the secret reference and replace it with actual value.
+func processMapForSecrets(config map[string]interface{}, secretManagerClient *secretmanager.SecretManagerClient) error {
+	for key, value := range config {
+		switch v := value.(type) {
+		case map[string]interface{}:
+			if err := processMapForSecrets(v, secretManagerClient); err != nil {
+				return fmt.Errorf("error processing map for key %s: %w", key, err)
+			}
+		case string:
+			if strings.HasPrefix(v, "${") && strings.HasSuffix(v, "}") {
```
kklimonda-fn, line 270: I wonder if this shouldn't be a regexp, instead of assuming that the secret must be a full string. I could imagine that we'd like to allow users to specify HTTP headers like: `X-Authorization: Bearer ${secret_manager.bearer_token}`

pkg/otelcollector/infra-meter/infra-meter.go
```diff
@@ -231,3 +257,68 @@ func updateK8sAttributesProcessor(processor interface{}, selectorsList []interfa
...
+	secretName = strings.TrimSuffix(secretName, "}")
+
+	secretName = fmt.Sprintf("projects/%s/secrets/%s-%s/versions/latest", secretManagerClient.GCPProjectID, secretManagerClient.ProjectID, secretName)
+
+	accessRequest := &secretmanagerpb.AccessSecretVersionRequest{
+		Name: secretName,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
kklimonda-fn, line 313: Do we need this defer here? By the time we're out of the function the request will be finished (failed or success)

pkg/secretmanager/provide.go
```diff
@@ -0,0 +1,81 @@
...
+func provideSecretManagerClient(in SecretManagerClientIn) (*SecretManagerClient, error) {
+	if in.InstallationModeConfig.InstallationMode != utils.InstallationModeCloudAgent {
+		return nil, nil
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	client, err := secretmanager.NewClient(ctx)
+	if err != nil {
+		defer cancel()
```
kklimonda-fn, line 48: If we defer cancel (see above whether we should) we probably want to do this once anyway?

kklimonda-fn (approved):

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: I think we need this here as a Golang practise for context.WithTimeout when the request is completed or timed out as this is a separate context for making calls to GCP

pkg/secretmanager/provide.go
hdkshingala: Here, the cancel has to be called either in OnStop when the agent is stopping or when there is some errors. We can't defer it once as it would cancel the context even before the onStart is executed and we would be using cancelled context after OnStart

pkg/otelcollector/infra-meter/infra-meter.go
hdkshingala: I followed the OTEL's way of reading the environment variables as I don't think they support combining string with variable. I can check again

PR #2934, opened 2023-11-15
Refactor aperture-go

hasit:
### Description of change
- Refactor multiple files
- Add protobuf files in gen/proto/flowcontrol dir
- Add java proto gen files

##### Checklist

- [ ] Breaking changes


sdks/aperture-go/example/main.go
```diff
@@ -63,16 +62,20 @@ func grpcOptions(insecureMode, skipVerify bool) []grpc.DialOption {
 func main() {
 	ctx := context.Background()
 
-	stdr.SetVerbosity(2)
+	apertureAgentAddr := getEnvOrDefault("APERTURE_AGENT_ADDRESS", defaultAgentAddress)
```
tanveergill, line 65: don't want to get anything through env. supposed to be provided explicitly by the developer. check typescript sdk

hasit: This is example code.

PR #3009, opened 2023-12-01
flow methods for response code and wait

sahil-lakhwani:
##### Checklist

- [ ] Breaking changes


sdks/aperture-go/sdk/flow.go
```diff
@@ -46,6 +46,8 @@ type Flow interface {
 	Span() trace.Span
 	End() error
 	CheckResponse() *checkv1.CheckResponse
+	RetryAfter() time.Duration
+	HTTPResponseCode() checkv1.StatusCode
```
tanveergill, line 50: How can the user check which status code it is against this? Should we simply return an int for simplicity?
With this approach, the user will have to import our proto to check the status code, right?

sahil-lakhwani: I think have correct types help in keeping the consistency. And I see `CheckResponse() *checkv1.CheckResponse` already asks to import our proto

tanveergill: @sahil-lakhwani,

I know we have check response. But we want to move away from such raw APIs and provide a user friendly APIs for commonly needed functionality

sahil-lakhwani: Also returning 200 for StatusCode Empty in all SDKs, as that is the default, for example in the case of no policies

tanveergill (approved):

PR #3070, opened 2023-12-22
Add the new FlowEnd call in the SDKs

hdkshingala:
##### Checklist


sdks/aperture-go/sdk/client.go
```diff
@@ -51,6 +51,8 @@ type FlowParams struct {
 	ResultCacheKey string
 	// GlobalCacheKeys are keys to global cache entries that need to be fetched at flow start.
 	GlobalCacheKeys []string
+	// If ExpectEnd is set to true, then the end of the request is expected.
+	ExpectEnd bool
```
tanveergill, line 55: This does not come from the user. Supposed to be hardcoded to true for integrations where we make an explicit FlowEnd call

hdkshingala: Okay.. Is it okay when we don't do flow end call due to no limiter decisions in the check response?

tanveergill: Yes, in fact in aperture-js we don't even make the flow end call if there are no limiter decisions of type ConcurrencyLimiter or ConcurrencyScheduler

hdkshingala: Ah right, I miss read that part. Will make the changes

hdkshingala: Handled it in https://github.com/fluxninja/aperture/pull/3070/commits/b2d399e660bb539609ee9297bd9cbcdf83145814#diff-12a8ea62371a6b3d57bca501dc84b627382e5a21127be2dbb5cf1122a7a791a9

tanveergill (approved):

